PR #8, opened 2021-04-20
Publish to maven local and add CI. 

dblock:
*Issue #, if available:*

*Description of changes:*

- Enable CI on pull requests and all other pushes
- Publish to maven Local
 
*CheckList:*

[x] Commits are signed per the DCO using --signoff

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -80,7 +80,7 @@ task javadocJar(type: Jar) {
     from javadoc.destinationDir
 }
 
-version '1.15.0.0'
+version '1.0.0.0'
 
 publishing {
     publications {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -100,6 +80,8 @@ task javadocJar(type: Jar) {
     from javadoc.destinationDir
 }
 
+version '1.15.0.0'
```
saratvemulapalli, line 83: I believe this should also be 1.0.0.0

saratvemulapalli (approved):
Changes look good to me. 
Thanks a lot for the change!

Just a minor comment.

build.gradle
dblock: Fixed.

Issue #19, opened 2021-05-18
Ensure common-utils can be a drop-in replacement for job-scheduler working in ODFE 1.13 and ES 7.10.x for v1.0.0


dblock:
Coming from [opensearch-plugins#12](https://github.com/opensearch-project/opensearch-plugins/issues/12), enable upgrade to OpenSearch, while supporting backward compatibility.

- [x] Rename Namespaces (e.g. `com.amazon.opendistroforelasticsearch` to `org.opensearch`)
- [x] Rename Classes, Methods, Variables 
- [x] Rename Remaining Identifiers (e.g. `Opendistro` to `OpenSearch`).
- [ ] Rename, and Backwards Compatibility for Rest APIs
- [ ] Rename, and Backwards Compatibility for Settings 
- [ ] Rename, and Backwards Compatibility for Indices
- [ ] Run in a backwards compatible way on top of OpenSearch 1.0 that has joined an ES 7.10.x cluster 
- [ ] Run in a backwards compatible way on top of OpenSearch 1.0 that has joined an ODFE 1.13.x cluster 
- [ ] Drop in replacement for the Opendistro version 1.13 of the plugin

VachaShah:
Renaming of the namespaces, classes, variables, methods and identifiers is done to reflect OpenSearch. There are some settings and APIs being used from security which need to be updated once security-plugin comes in with the update. I have opened an issue for this: #22. 

For the above checklist, the remaining part is the update in `CI.yml` to required OpenSearch version.

dblock:
@VachaShah What else needs to be done for this, or can it be closed?

VachaShah:
@dblock Looks like all the required items are complete for this, it can be closed.

PR #21, opened 2021-05-18
Rename classes, variables, methods to incorporate OpenSearch

VachaShah:
Signed-off-by: Vacha Shah <vachshah@amazon.com>

### Description
This PR updates the classes, variables, methods to incorporate OpenSearch.
 
### Issues Resolved
Part 2 of #19 
 
### Check List
  - [X] All tests pass


src/main/java/org/opensearch/commons/authuser/AuthUserRequestBuilder.java
```diff
@@ -42,7 +42,7 @@ public AuthUserRequestBuilder(String auth) {
     }
 
     public Request build() {
-        Request request = new Request("GET", "/_opendistro/_security/authinfo");
+        Request request = new Request("GET", "/_opensearch/_security/authinfo");
```
dblock, line 45: You need to track this down in the security plugin and make sure that this is the change that was made there. I think the path is different.

dblock (changes requested):
Similar to the example below, renaming actual paths and settings needs to be done carefully. Please check names? Thanks.

src/main/java/org/opensearch/commons/authuser/AuthUserRequestBuilder.java
dblock: Please open an issue in this repo to change this later, linking to an issue in security to watch.

dblock (approved):

PR #24, opened 2021-05-21
Version changed to rc1

skkosuri-amzn:
### Description
Rename artifacts version to rc1.
 
### Issues Resolved
Rename artifacts version to rc1.
 
### Check List
  - [x] All tests pass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@
 buildscript {
     ext {
         opensearch_group = "org.opensearch"
-        opensearch_version = System.getProperty("opensearch.version", "1.0.0-beta1")
+        opensearch_version = System.getProperty("opensearch.version", "1.0.0-rc1")
     }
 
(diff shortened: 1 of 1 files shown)


aditjind (approved):

bowenlan-amzn:
leave a note, some point we need to update the workflow file to the new version also

dblock:
Don't we need to do this all at the same time, starting with OpenSearch? Otherwise this change breaks all plugins that depend on it?

dblock (approved):
I'll hit approve cause by itself the change is fine, but see my comment.

dblock:
Will you also change beta1 to rc1 in https://github.com/opensearch-project/common-utils/blob/main/.github/workflows/ci.yml? It doesn't really matter for the build which identifiers are used, but this way we know we're there.

PR #38, opened 2021-07-14
Fix class loader issue for notifications response

### LINKED ISSUE #37: [BUG] Calling Notifications from other plugins results in ClassLoader error
**Describe the bug**
https://github.com/opensearch-project/common-utils/blob/4d8fb67fa99d87e40de5180453d7c894dd70de6f/src/main/kotlin/org/opensearch/commons/notifications/NotificationsPluginInterface.kt#L199
Tested in reporting plugin. When reporting calls this function, the following error will show up.
```
java.lang.ClassCastException: class org.opensearch.commons.notifications.action.SendNotificationRequest cannot be cast to class org.opensearch.commons.notifications.action.SendNotificationRequest (org.opensearch.commons.notifications.action.SendNotificationRequest is in unnamed module of loader java.net.FactoryURLClassLoader @1229a2b7; org.opensearch.commons.notifications.action.SendNotificationRequest is in unnamed module of loader java.net.FactoryURLClassLoader @28f9fedd)
        at org.opensearch.notifications.action.SendNotificationAction.doExecute(SendNotificationAction.kt:47) ~[?:?]
        at org.opensearch.action.support.TransportAction$RequestFilterChain.proceed(TransportAction.java:192) [opensearch-1.0.0-rc1.jar:1.0.0-rc1]
        at org.opensearch.indexmanagement.rollup.actionfilter.FieldCapsFilter.apply(FieldCapsFilter.kt:136) [opensearch-index-management-1.0.0.0-rc1.jar:1.0.0.0-rc1]
        at org.opensearch.action.support.TransportAction$RequestFilterChain.proceed(TransportAction.java:190) [opensearch-1.0.0-rc1.jar:1.0.0-rc1]
...
```

**To Reproduce**

**Expected behavior**

**Plugins**
Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**

**Additional context**

joshuali925:
Signed-off-by: Joshua Li <joshuali925@gmail.com>

### Description
fix from @dai-chen:
Recreate the response object from `ActionResponse` as specific notifications response in notifications plugin interface to avoid classloader issue when passing response between plugins
 
### Issues Resolved
#37 
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/notifications/NotificationsPluginInterface.kt
+++ b/src/main/kotlin/org/opensearch/commons/notifications/NotificationsPluginInterface.kt
@@ -72,7 +72,6 @@ object NotificationsPluginInterface {
      * @param request The request object
      * @param listener The listener for getting response
      */
-    @Suppress("UNCHECKED_CAST")
     fun createNotificationConfig(
         client: NodeClient,
         request: CreateNotificationConfigRequest,
@@ -91,7 +90,6 @@ object NotificationsPluginInterface {
      * @param request The request object
      * @param listener The listener for getting response
      */
-    @Suppress("UNCHECKED_CAST")
     fun updateNotificationConfig(
         client: NodeClient,
         request: UpdateNotificationConfigRequest,
@@ -110,7 +108,6 @@ object NotificationsPluginInterface {
      * @param request The request object
      * @param listener The listener for getting response
      */
-    @Suppress("UNCHECKED_CAST")
     fun deleteNotificationConfig(
         client: NodeClient,
         request: DeleteNotificationConfigRequest,
@@ -129,7 +126,6 @@ object NotificationsPluginInterface {
      * @param request The request object
      * @param listener The listener for getting response
      */
-    @Suppress("UNCHECKED_CAST")
     fun getNotificationConfig(
         client: NodeClient,
         request: GetNotificationConfigRequest,
@@ -148,7 +144,6 @@ object NotificationsPluginInterface {
      * @param request The request object
      * @param listener The listener for getting response
      */
-    @Suppress("UNCHECKED_CAST")
     fun getNotificationEvent(
         client: NodeClient,
         request: GetNotificationEventRequest,
@@ -167,7 +162,6 @@ object NotificationsPluginInterface {
      * @param request The request object
      * @param listener The listener for getting response
      */
-    @Suppress("UNCHECKED_CAST")
     fun getPluginFeatures(
         client: NodeClient,
         request: GetPluginFeaturesRequest,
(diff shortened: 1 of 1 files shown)


dai-chen (approved):
Thanks for the fix!

dblock (commented):
I love this as a start.

Try to refactor this where you don't need to `override fun onResponse` manually in every instance. 

I think I want to be able to write this:

```java
wrapper.execute(
    SEND_NOTIFICATION_ACTION_TYPE,
    SendNotificationRequest(eventSource, channelMessage, channelIds, threadContext),
    object: SomeNewType<ActionResponse, SendNotificationResponse>
}
```

dblock (changes requested):
This should be refactored, and definitely needs tests to be merged.

dai-chen:
> I love this as a start.
> 
> Try to refactor this where you don't need to `override fun onResponse` manually in every instance.
> 
> I think I want to be able to write this:
> 
> ```java
> wrapper.execute(
>     SEND_NOTIFICATION_ACTION_TYPE,
>     SendNotificationRequest(eventSource, channelMessage, channelIds, threadContext),
>     object: SomeNewType<ActionResponse, SendNotificationResponse>
> }
> ```

I was trying to extract a `wrapActionListener` method though had problem with `recreateObject` which requires reified type rather than generic type. Let me give it another try.

dblock:
@joshuali925 excellent!

Is it worth putting this into a method that takes `DELETE_NOTIFICATION_CONFIG_ACTION_TYPE`, and `DeleteNotificationConfigResponse` or are we trying too hard?

```java
client.execute(
            DELETE_NOTIFICATION_CONFIG_ACTION_TYPE,
            request,
            listener
            wrapActionListener(listener) { response -> recreateObject(response) { DeleteNotificationConfigResponse(it) } }
        )
```

joshuali925:
@dblock agree that could look cleaner, but in `sendNotification` the client is wrapped with `SecureClientWrapper` before calling `execute`. I assume we'll need to make this an exception if we generalize the execute function, and i'm not sure if there will be other exceptions in the future. Anyways I'll also defer to @dai-chen since he is coming up with the fixes and refactoring
https://github.com/opensearch-project/common-utils/blob/d5bcf10798632556e548442221e3c42936264273/src/main/kotlin/org/opensearch/commons/notifications/NotificationsPluginInterface.kt#L218-L225

dai-chen:
Closing for https://github.com/opensearch-project/common-utils/pull/40

dai-chen:
> @dblock agree that could look cleaner, but in `sendNotification` the client is wrapped with `SecureClientWrapper` before calling `execute`. I assume we'll need to make this an exception if we generalize the execute function, and i'm not sure if there will be other exceptions in the future. Anyways I'll also defer to @dai-chen since he is coming up with the fixes and refactoring
> https://github.com/opensearch-project/common-utils/blob/d5bcf10798632556e548442221e3c42936264273/src/main/kotlin/org/opensearch/commons/notifications/NotificationsPluginInterface.kt#L218-L225

@joshuali925 @dblock I tried this idea but the code was not clean as I thought. This is mostly because of generic types and number fo params required by each execute call. I would suggest keeping it as is for now.

PR #40, opened 2021-07-19
Fix class loader issue for notifications response

dai-chen:
### Description
Type cast error occurred when entering `ActionResponse.onResponse()`. The reason is due to different class loaders across plugins that explained in details here: https://github.com/opensearch-project/notifications/pull/223. To solve the problem in similar way, wrap the listener by a new one with only difference on generic type (which is abstract `ActionResponse` rather than concrete `XXXNotificationResponse` to delay the casting operation after recreating response object by class loaded by current plugin class loader.

The main wrapper logic is patched from this https://github.com/opensearch-project/common-utils/pull/38. This PR is focused on adding missing UT to cover all notification APIs touched.
 
### Issues Resolved
https://github.com/opensearch-project/common-utils/issues/37
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/notifications/NotificationsPluginInterface.kt
+++ b/src/main/kotlin/org/opensearch/commons/notifications/NotificationsPluginInterface.kt
@@ -72,7 +72,6 @@ object NotificationsPluginInterface {
      * @param request The request object
      * @param listener The listener for getting response
      */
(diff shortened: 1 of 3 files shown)


dblock (approved):
👏

joshuali925 (approved):

Issue #44, opened 2021-08-03
Support ISM publishing notifications to notification plugin with pass through API


dbbaughe:
**Is your feature request related to a problem? Please describe.**
https://github.com/opensearch-project/notifications/issues/253

Adding the legacy classes/request/response/etc. needed for ISM to publish legacy notifications.

getsaurabh02:
With https://github.com/opensearch-project/common-utils/pull/53 being merged, don't see any pending action items here.

PR #45, opened 2021-08-03
Adds legacy chime, slack, custom webhook messages, request/response f…

dbbaughe:
…or publishing legacy notifications, and method for executing transport action

Signed-off-by: Drew Baugher <46505179+dbbaughe@users.noreply.github.com>

### Description
Adds required legacy messages, request/response classes, and methods for ISM to publish a notification through the notification plugin with the old embedded types without having to create a channel first.
opensearch-project/notifications#253
 
### Issues Resolved
https://github.com/opensearch-project/common-utils/issues/44

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #53, opened 2021-08-11
Adds legacy chime, slack, custom webhook messages, request/response f

dbbaughe:
…or publishing legacy notifications, and method for executing transport action

Signed-off-by: Drew Baugher 46505179+dbbaughe@users.noreply.github.com

### Description
Adds required legacy messages, request/response classes, and methods for ISM to publish a notification through the notification plugin with the old embedded types without having to create a channel first.
opensearch-project/notifications#253
 
### Issues Resolved
https://github.com/opensearch-project/common-utils/issues/44
 
### Check List
  - [x] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added

### DIFF
--- a/src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
+++ b/src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
@@ -1,0 +1,127 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package org.opensearch.commons.destination.message;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Map;
+
+import org.apache.http.client.utils.URIBuilder;
+import org.opensearch.common.Strings;
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.common.io.stream.Writeable;
+
+/**
+ * This class holds the generic parameters required for a
+ * message.
+ */
+public abstract class LegacyBaseMessage implements Writeable {
+
+    protected LegacyDestinationType destinationType;
+    protected String destinationName;
+    protected String url;
+    private String content;
+
+    LegacyBaseMessage(final LegacyDestinationType destinationType, final String destinationName, final String content) {
+        if (destinationType == null) {
+            throw new IllegalArgumentException("Channel type must be defined");
+        }
+        if (!Strings.hasLength(destinationName)) {
+            throw new IllegalArgumentException("Channel name must be defined");
+        }
+        this.destinationType = destinationType;
+        this.destinationName = destinationName;
+        this.content = content;
+    }
+
+    LegacyBaseMessage(final LegacyDestinationType destinationType, final String destinationName, final String content, final String url) {
+        this(destinationType, destinationName, content);
+        if (url == null) {
+            throw new IllegalArgumentException("url is invalid or empty");
+        }
+        this.url = url;
+    }
+
+    LegacyBaseMessage(org.opensearch.common.io.stream.StreamInput streamInput) throws IOException {
+        this.destinationType = streamInput.readEnum(LegacyDestinationType.class);
+        this.destinationName = streamInput.readString();
+        this.url = streamInput.readOptionalString();
+        this.content = streamInput.readString();
+    }
+
+    public void setUrl(String url) {
+        this.url = url;
+    }
+
+    public LegacyDestinationType getChannelType() {
+        return destinationType;
+    }
+
+    public String getChannelName() {
+        return destinationName;
+    }
+
+    public String getMessageContent() {
+        return content;
+    }
+
+    public String getUrl() {
+        return url;
+    }
+
+    public URI getUri() {
+        return buildUri(getUrl().trim(), null, null, -1, null, null);
+    }
+
+    protected URI buildUri(String endpoint, String scheme, String host, int port, String path, Map<String, String> queryParams) {
+        try {
+            if (Strings.isNullOrEmpty(endpoint)) {
+                if (Strings.isNullOrEmpty(scheme)) {
+                    scheme = "https";
+                }
+                URIBuilder uriBuilder = new URIBuilder();
+                if (queryParams != null) {
+                    for (Map.Entry<String, String> e : queryParams.entrySet())
+                        uriBuilder.addParameter(e.getKey(), e.getValue());
+                }
+                return uriBuilder.setScheme(scheme).setHost(host).setPort(port).setPath(path).build();
+            }
+            return new URIBuilder(endpoint).build();
+        } catch (URISyntaxException exception) {
+            throw new IllegalStateException("Error creating URI");
+        }
+    }
+
+    @Override
+    public void writeTo(StreamOutput streamOutput) throws IOException {
+        streamOutput.writeEnum(destinationType);
+        streamOutput.writeString(destinationName);
+        streamOutput.writeOptionalString(url);
+        streamOutput.writeString(content);
+    }
+}
--- a/src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
+++ b/src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
@@ -43,7 +43,7 @@ import org.opensearch.common.io.stream.Writeable;
  */
 public abstract class LegacyBaseMessage implements Writeable {
 
-    protected LegacyDestinationType destinationType;
+    private final LegacyDestinationType destinationType;
     protected String destinationName;
     protected String url;
     private String content;
--- a/src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
+++ b/src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
@@ -33,6 +33,7 @@ import java.util.Map;
 
 import org.apache.http.client.utils.URIBuilder;
 import org.opensearch.common.Strings;
+import org.opensearch.common.io.stream.StreamInput;
 import org.opensearch.common.io.stream.StreamOutput;
 import org.opensearch.common.io.stream.Writeable;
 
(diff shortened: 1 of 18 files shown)


lezzago (dismissed):

akbhatta (changes requested):
I see same code in the PR as well. please refer other PR for my comments

dbbaughe:
@akbhatta As mentioned on previous PR, most of this code is ported over from alerting and the intention is to keep it the same, i.e. not converting it to Kotlin and not changing business logic. Reason being this is only for ISM, ISM was already using these classes as they were, no other plugin will be using these, and we want to ensure they work the same as they did previously. This PR is simply to bring them over, if we decide to refactor them we'll do it in a later PR.

akbhatta:
> @akbhatta As mentioned on previous PR, most of this code is ported over from alerting and the intention is to keep it the same, i.e. not converting it to Kotlin and not changing business logic. Reason being this is only for ISM, ISM was already using these classes as they were, no other plugin will be using these, and we want to ensure they work the same as they did previously. This PR is simply to bring them over, if we decide to refactor them we'll do it in a later PR.

Even though the code might be just moving over, the environment it will be used will be completely different. I would recommend fixing the issues before merging.

dbbaughe:
> > @akbhatta As mentioned on previous PR, most of this code is ported over from alerting and the intention is to keep it the same, i.e. not converting it to Kotlin and not changing business logic. Reason being this is only for ISM, ISM was already using these classes as they were, no other plugin will be using these, and we want to ensure they work the same as they did previously. This PR is simply to bring them over, if we decide to refactor them we'll do it in a later PR.
> 
> Even though the code might be just moving over, the environment it will be used will be completely different. I would recommend fixing the issues before merging.

Can you clarify on this part -> "the environment it will be used will be completely different"? This keeps the usage in ISM the same w/ no code changes. The classes were originally written in Java so there should be no issues there. The only new part is the fact that the notification Kotlin project is utilizing them specifically in the one transport action for the pass through API. But, no matter what new code has to be written somewhere and since that is all new code to being with it makes more sense to have it there rather than make changes in ISM for the backwards compatible support of old destinations.

akbhatta:
> > > @akbhatta As mentioned on previous PR, most of this code is ported over from alerting and the intention is to keep it the same, i.e. not converting it to Kotlin and not changing business logic. Reason being this is only for ISM, ISM was already using these classes as they were, no other plugin will be using these, and we want to ensure they work the same as they did previously. This PR is simply to bring them over, if we decide to refactor them we'll do it in a later PR.
> > 
> > 
> > Even though the code might be just moving over, the environment it will be used will be completely different. I would recommend fixing the issues before merging.
> 
> Can you clarify on this part -> "the environment it will be used will be completely different"? This keeps the usage in ISM the same w/ no code changes. The classes were originally written in Java so there should be no issues there. The only new part is the fact that the notification Kotlin project is utilizing them specifically in the one transport action for the pass through API. But, no matter what new code has to be written somewhere and since that is all new code to begin with it makes more sense to have it there rather than make changes in ISM for the backwards compatible support of old destinations.

Even though ISM code is almost same (I believe it cannot be exact same), it will be transport call instead of internal java call. the same classes are used in notification plugin as well as you mentioned. If there are issues with the code previously uncovered, it might cause problem in notification plugin or in transport layer.  You need not do any refactor (or convert to kotlin if you are not comfortable) however please address the issues in the code.

dbbaughe:
@akbhatta Of the comments on old PR, username/password were removed, and the main other one I see is regarding the constructing of the URI from parts being done by caller. I can make that change, were there any others that you were concerned about?

src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
```diff
@@ -0,0 +1,128 @@
...
+import org.opensearch.common.Strings;
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.common.io.stream.Writeable;
+
+/**
+ * This class holds the generic parameters required for a
+ * message.
+ */
+public abstract class LegacyBaseMessage implements Writeable {
```
akbhatta, line 44: Can you please make the data class immutable. It is easier to maintain lifecycle of mutable object.

src/main/java/org/opensearch/commons/destination/message/LegacyChimeMessage.java
```diff
@@ -0,0 +1,99 @@
...
+    private LegacyChimeMessage(
+        final LegacyDestinationType destinationType,
+        final String destinationName,
+        final String url,
+        final String message
+    ) {
+
+        super(destinationType, destinationName, message, url);
+
+        if (LegacyDestinationType.CHIME != destinationType) {
```
akbhatta, line 49: Since this class is only for chime, can this value be inferred instead of passing from the caller?

src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
```diff
@@ -0,0 +1,128 @@
...
+
+    protected LegacyDestinationType destinationType;
+    protected String destinationName;
+    protected String url;
+    private String content;
```
akbhatta, line 49: Why is the message and the destination are clubbed into single class, can you please separate it?

src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
```diff
@@ -0,0 +1,128 @@
...
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.common.io.stream.Writeable;
+
+/**
+ * This class holds the generic parameters required for a
+ * message.
+ */
+public abstract class LegacyBaseMessage implements Writeable {
+
+    protected LegacyDestinationType destinationType;
```
akbhatta, line 46: since the getters are available, can this members be private?

src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
```diff
@@ -0,0 +1,128 @@
...
+
+    public String getMessageContent() {
+        return content;
+    }
+
+    public String getUrl() {
+        return url;
+    }
+
+    public URI getUri() {
```
akbhatta, line 98: why are we using both URI and URL terms?

src/main/java/org/opensearch/commons/destination/message/LegacyChimeMessage.java
```diff
@@ -0,0 +1,99 @@
...
+        this.message = message;
+    }
+
+    public LegacyChimeMessage(StreamInput streamInput) throws IOException {
+        super(streamInput);
+        this.message = super.getMessageContent();
+    }
+
+    @Override
+    public String toString() {
```
akbhatta, line 66: since we are using message in toString, how are we making sure that this object is not used for logging anywhere?

src/main/java/org/opensearch/commons/destination/message/LegacyCustomWebhookMessage.java
```diff
@@ -0,0 +1,289 @@
...
+import org.opensearch.common.Strings;
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+
+/**
+ * This class holds the content of an CustomWebhook message
+ */
+public class LegacyCustomWebhookMessage extends LegacyBaseMessage {
+
+    private final String message;
```
akbhatta, line 45: can we make use of buildUri() and only save 1 object (uri or url) ?

src/main/java/org/opensearch/commons/destination/message/LegacyDestinationType.java
```diff
@@ -0,0 +1,36 @@
...
+ * permissions and limitations under the License.
+ */
+
+package org.opensearch.commons.destination.message;
+
+/**
+ * Supported legacy notification destinations for Index Management
+ */
+public enum LegacyDestinationType {
+    CHIME,
```
akbhatta, line 33: Importing might be problem if we use the values same. can we rename it to LEGACY_CHIME etc?
Alternative is to use Config type and have a check enabling only Slack, Chime and webhook.

src/main/kotlin/org/opensearch/commons/notifications/NotificationsPluginInterface.kt
```diff
@@ -217,6 +221,30 @@ object NotificationsPluginInterface {
...
+    ) {
+        if (request.feature != Feature.INDEX_MANAGEMENT) {
+            // Do not change this; do not pass in Feature.INDEX_MANAGEMENT if you are not the Index Management plugin.
+            throw IllegalArgumentException("The publish notification method only supports the Index Management feature.")
+        }
+
+        client.execute(
+            LEGACY_PUBLISH_NOTIFICATION_ACTION_TYPE,
+            request,
+            wrapActionListener(listener) { response -> recreateObject(response) { LegacyPublishNotificationResponse(it) } }
```
akbhatta, line 244: Please let me know if recreateObject is required here. we may have to update in other transport calls as well. (as of now it seems it works without it)

src/main/kotlin/org/opensearch/commons/notifications/action/LegacyPublishNotificationRequest.kt
```diff
@@ -0,0 +1,83 @@
...
+        super.writeTo(output)
+        output.writeEnum(baseMessage.channelType)
+        baseMessage.writeTo(output)
+        output.writeEnum(feature)
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    override fun validate(): ActionRequestValidationException? = null
```
akbhatta, line 82: if no validation is required can we remove overriding this function?

src/main/kotlin/org/opensearch/commons/notifications/action/LegacyPublishNotificationResponse.kt
```diff
@@ -0,0 +1,63 @@
...
+     * {@inheritDoc}
+     */
+    @Throws(IOException::class)
+    override fun writeTo(output: StreamOutput) {
+        destinationResponse.writeTo(output)
+    }
+
+    // This class is only used across transport wire and does not need to implement toXContent
+    override fun toXContent(builder: XContentBuilder, params: ToXContent.Params): XContentBuilder {
+        return builder
```
akbhatta, line 61: Please throw exception since this method is not expected to be called.

akbhatta (changes requested):

akbhatta:
other comments mentioned in PR #45 can be punted to next PR when you are refactoring the code.

src/main/kotlin/org/opensearch/commons/notifications/action/LegacyPublishNotificationRequest.kt
dbbaughe: No, it's an abstract method.

src/main/kotlin/org/opensearch/commons/notifications/NotificationsPluginInterface.kt
dbbaughe: If I remember previously, this was required for class loader issue. All of the transport calls are following this way as defined in rest of fail.

src/main/java/org/opensearch/commons/destination/message/LegacyChimeMessage.java
dbbaughe: This is legacy code from alerting release, logic isn't changing in this PR. I'd presume w/ code reviews and pen testing to make sure it doesn't show up in logs. Either way doesn't matter, will obfuscate the message.

src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
dbbaughe: Legacy code from Alerting, @vamshin if you remember why you used both feel free to chime in.

src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
dbbaughe: Legacy from Alerting implementation, not looking to break apart existing classes in this PR.

src/main/java/org/opensearch/commons/destination/message/LegacyBaseMessage.java
dbbaughe: Can be done in a refactoring PR later on, for existing code that is just being moved to this repository keeping changes to a minimum.

lezzago (approved):

PR #116, opened 2022-02-17
Using Github App token to trigger CI for version increment PRs

VachaShah:
Signed-off-by: Vacha Shah <vachshah@amazon.com>

### Description
Using Github App token to trigger CI on PRs created for version increment from the version.yml workflow. (https://github.com/opensearch-project/common-utils/pull/111#issuecomment-1039635451)
 
### Issues Resolved
 
### Check List

### DIFF
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -8,9 +8,15 @@ on:
 jobs:  
   build:
     runs-on: ubuntu-latest
-    env:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
+      - name: GitHub App token
+        id: github_app_token
+        uses: tibdex/github-app-token@v1.5.0
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/version.yml
```diff
@@ -27,13 +33,17 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ref: ${{ env.BASE }}
+          token: ${{ steps.github_app_token.outputs.token }}
```
bowenlan-amzn, line 36: Curious, if we don't give the app token here, what things it could break, will it cause the new PR not triggering CI?

VachaShah: Do you mean for this specific step or overall for using the token?

bowenlan-amzn: Not overall, just for this checkout action. If this checkout still use the default token, we only pass this app token to the next create PR action, will it cause the new PR not triggering CI?

VachaShah: I think I have seen multiple recommendations from this GHA to use the same token in actions/checkout as well. One of them is https://github.com/peter-evans/create-pull-request/issues/990

dblock (approved):

PR #136, opened 2022-03-14
Remove feature and feature_list usage for Notifications

qreshi:
### Description
* Removes feature usage from NotificationConfig
* Removes feature usage from FeatureChannelList request
* Removes feature usage from EventSource
* Removes feature from LegacyPublishNotification

There will also be a complementary PR made in the Notifications project to integrate these changes.

There are still some indirect references to features like the enabled features returned in the GetPluginFeatures API which will be addressed in a follow-up PR.

### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/notifications/NotificationConstants.kt
+++ b/src/main/kotlin/org/opensearch/commons/notifications/NotificationConstants.kt
@@ -69,9 +69,5 @@ object NotificationConstants {
     const val ALLOWED_CONFIG_FEATURE_LIST_TAG = "allowed_config_feature_list"
     const val PLUGIN_FEATURES_TAG = "plugin_features"
 
-    const val FEATURE_ALERTING = "alerting"
-    const val FEATURE_INDEX_MANAGEMENT = "index_management"
-    const val FEATURE_REPORTS = "reports"
-
     const val DEFAULT_MAX_ITEMS = 1000
 }
--- a/src/main/kotlin/org/opensearch/commons/notifications/NotificationConstants.kt
+++ b/src/main/kotlin/org/opensearch/commons/notifications/NotificationConstants.kt
@@ -36,7 +36,6 @@ object NotificationConstants {
     const val NAME_TAG = "name"
     const val DESCRIPTION_TAG = "description"
     const val IS_ENABLED_TAG = "is_enabled"
-    const val FEATURE_LIST_TAG = "feature_list"
     const val TITLE_TAG = "title"
     const val SEVERITY_TAG = "severity"
     const val TAGS_TAG = "tags"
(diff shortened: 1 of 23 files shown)


src/main/kotlin/org/opensearch/commons/notifications/NotificationConstants.kt
```diff
@@ -36,7 +36,6 @@ object NotificationConstants {
     const val NAME_TAG = "name"
     const val DESCRIPTION_TAG = "description"
     const val IS_ENABLED_TAG = "is_enabled"
-    const val FEATURE_LIST_TAG = "feature_list"
```
lezzago, line 39: What about removing `FEATURE_CONFIG_LIST_TAG`?

lezzago:
Do we want to keep the [FeatureChannelList class](https://github.com/opensearch-project/common-utils/blob/bccb1bb329517c3c550df9b8fd9df81bbf05e4f1/src/main/kotlin/org/opensearch/commons/notifications/model/FeatureChannelList.kt)?

src/main/kotlin/org/opensearch/commons/notifications/NotificationConstants.kt
```diff
@@ -70,9 +69,5 @@ object NotificationConstants {
     const val ALLOWED_CONFIG_FEATURE_LIST_TAG = "allowed_config_feature_list"
     const val PLUGIN_FEATURES_TAG = "plugin_features"
```
dbbaughe, line 70: Also these other mentions of FEATURE? Are they still used?

src/main/kotlin/org/opensearch/commons/notifications/NotificationsPluginInterface.kt
```diff
@@ -211,11 +210,6 @@ object NotificationsPluginInterface {
         request: LegacyPublishNotificationRequest,
         listener: ActionListener<LegacyPublishNotificationResponse>
     ) {
-        if (request.feature != FEATURE_INDEX_MANAGEMENT) {
```
dbbaughe, line 214: Might want to update the description (e.g. "Internal API only for the Index Management plugin.")

src/main/kotlin/org/opensearch/commons/notifications/NotificationConstants.kt
qreshi: These all fall under the `"There are still some indirect references to features like the enabled features returned in the GetPluginFeatures API which will be addressed in a follow-up PR."` category.

I'll remove that then since that API is likely going to be taken out.

src/main/kotlin/org/opensearch/commons/notifications/NotificationConstants.kt
qreshi: Same as my comment to Drew.

qreshi:
> Do we want to keep the [FeatureChannelList class](https://github.com/opensearch-project/common-utils/blob/bccb1bb329517c3c550df9b8fd9df81bbf05e4f1/src/main/kotlin/org/opensearch/commons/notifications/model/FeatureChannelList.kt)?

It might be renamed later and the REST handler for Notifications might be modified but we do want to keep the function since listing Channels is a little different than just listing Notifications configs (it'd collapse more complicated channels like Email) so we'd want both.

lezzago (approved):

dbbaughe (approved):

PR #146, opened 2022-03-28
Change default opensearch version to 2.0.0-alpha1-SNAPSHOT

peterzhuamazon:
Signed-off-by: Peter Zhu <zhujiaxi@amazon.com>

### Description
Change default opensearch version to 2.0.0-alpha1-SNAPSHOT
 
### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-build/issues/1632
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@
 buildscript {
     ext {
         opensearch_group = "org.opensearch"
-        opensearch_version = System.getProperty("opensearch.version", "2.0.0-SNAPSHOT")
+        opensearch_version = System.getProperty("opensearch.version", "2.0.0-alpha1-SNAPSHOT")
         kotlin_version = System.getProperty("kotlin.version", "1.6.10")
     }
 


dbbaughe (approved):
Is this the long term solution for this? Or should it remain 2.0.0-SNAPSHOT and instead the build infrastructure should pass in the qualifier during the build of common utils so it can go and pull the alpha1 snapshots instead?

peterzhuamazon:
@dbbaughe this value is default means every new release you guys should change it for your checks.
As for infra we will pass the opensearch.version directly thus not an issue for us.

bbarani (approved):

dbbaughe:
@peterzhuamazon Yes, what I mean is we will need to now change this back in the future to 2.0.0-SNAPSHOT. If we had kept this as 2.0.0-SNAPSHOT instead, would it not have been possible for infra to pass the qualifier themselves during the build of this project instead of hardcoding the qualifier?

peterzhuamazon:
> @peterzhuamazon Yes, what I mean is we will need to now change this back in the future to 2.0.0-SNAPSHOT. If we had kept this as 2.0.0-SNAPSHOT instead, would it not have been possible for infra to pass the qualifier themselves during the build of this project instead of hardcoding the qualifier?

It would not block infra in anyway in our distribution pipeline if you keep 2.0.0-alpha1-SNAPSHOT as infra will pass `2.0.0` or `2.0.0-SNAPSHOT` in infra jenkins anyway.

dbbaughe:
> > @peterzhuamazon Yes, what I mean is we will need to now change this back in the future to 2.0.0-SNAPSHOT. If we had kept this as 2.0.0-SNAPSHOT instead, would it not have been possible for infra to pass the qualifier themselves during the build of this project instead of hardcoding the qualifier?
> 
> It would not block infra in anyway in our distribution pipeline if you keep 2.0.0-alpha1-SNAPSHOT as infra will pass `2.0.0` or `2.0.0-SNAPSHOT` in infra jenkins anyway.
> 
> We will override -Dopensearch.version param in our build.

Got it, so sounds like this change wasn't really needed then?

peterzhuamazon:
> > > @peterzhuamazon Yes, what I mean is we will need to now change this back in the future to 2.0.0-SNAPSHOT. If we had kept this as 2.0.0-SNAPSHOT instead, would it not have been possible for infra to pass the qualifier themselves during the build of this project instead of hardcoding the qualifier?
> > 
> > 
> > It would not block infra in anyway in our distribution pipeline if you keep 2.0.0-alpha1-SNAPSHOT as infra will pass `2.0.0` or `2.0.0-SNAPSHOT` in infra jenkins anyway.
> > We will override -Dopensearch.version param in our build.
> 
> Got it, so sounds like this change wasn't really needed then?

It is needed, because your CI is still pointing to the weeks old 2.0.0-SNAPSHOT instead of 2.0.0-alpha1-SNAPSHOT.

peterzhuamazon:
It is possible when we reach 2.0.0GA there are new core changes so suggest you still change default to 2.0.0-SNAPSHOT when it comes for your CI to be pointed to correct snapshot.

PR #155, opened 2022-04-04
add SQL/PPL transport request/response models

zhongnansu:
Signed-off-by: Zhongnan Su <szhongna@amazon.com>

### Description
- Add the transport interfaces for SQL plugin, to enable sending PPL/SQL query between other plugin in transport calls
- Define corresponding request/response data models that any plugin can consume
- [Related PR in SQL plugin](https://github.com/opensearch-project/sql/pull/533) won't pass CI, until this PR is merged and artifacts is released to central repo.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


dblock:
@zhongnansu This looks old and unattended. Do you still want this merged? I'll find someone to CR.

getsaurabh02:
@zhongnansu Are you still blocked on review for this PR?

zhongnansu:
@dblock @getsaurabh02 Yes, I need to merge this in. At that time I don't want it to affect 2.0 release. Now since 2.0 is cut, I can contribute this feature. will update the PR today and re-request review, thx

joshuali925 (dismissed):

src/main/kotlin/org/opensearch/commons/sql/action/TransportQueryResponse.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.opensearch.common.io.stream.Writeable
+import org.opensearch.common.xcontent.ToXContent
+import org.opensearch.common.xcontent.XContentBuilder
+import java.io.IOException
+
+/**
+ * Action Response for ppl and sql query.
+ */
+class TransportQueryResponse : BaseResponse {
+    val queryResponse: String
```
penghuo, line 19: Instead of return plain string, we talked about DataFrame/DataSet. Please create an issue to describe the idea and link to this PR.

src/main/kotlin/org/opensearch/commons/sql/action/SQLActions.kt
```diff
@@ -0,0 +1,17 @@
+package org.opensearch.commons.sql.action
+
+import org.opensearch.action.ActionType
+
+object SQLActions {
```
penghuo, line 5: How about PPL? reuse this action?

penghuo:
I have concern regarding to maintenance the transport layer interface. Ideally, SQL/PPL should be a module of Opensearch, then other application plugins could directly use it. Could we consider move Transport interface to SQL/PPL for now. Then we can decide later how we want to separate them when we have real use case?

praveensameneni:
> Ideally, SQL/PPL should be a module of Opensearch, then other application plugins could directly use it

I like this idea. @penghuo , can you create an issue on core and let's discuss if we can move this to modules so other plugins can directly use it. What do you think could be the changes needed for other plugins to use SQL/PPL without much modifications.

zhongnansu:
> I have concern regarding to maintenance the transport layer interface. Ideally, SQL/PPL should be a module of Opensearch, then other application plugins could directly use it. Could we consider move Transport interface to SQL/PPL for now. Then we can decide later how we want to separate them when we have real use case?

@penghuo @praveensameneni Thanks for the feedback. Separating SQL/PPL codebase into 2 repos without doesn't seem like a long term maintainable solution. Besides, like your previous comment mentioned,  without a response wrapper, the client will find it difficult to use just pure json. I'll close this PR and link to your new issue

PR #158, opened 2022-04-13
Support sending email message via Notifications passthrough API

qreshi:
### Description
The Notifications passthrough API is used for sending legacy messages for consumers of the old notification subproject in Alerting. Alerting has support for Email Destinations that need to use the passthrough API as a fallback mechanism for Destinations that have not migrated.

This PR adds support for email via passthrough.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/notifications/model/NotificationEvent.kt
+++ b/src/main/kotlin/org/opensearch/commons/notifications/model/NotificationEvent.kt
@@ -4,7 +4,6 @@
  */
 package org.opensearch.commons.notifications.model
 
-import org.opensearch.OpenSearchException
 import org.opensearch.common.io.stream.StreamInput
 import org.opensearch.common.io.stream.StreamOutput
 import org.opensearch.common.io.stream.Writeable
@@ -112,7 +111,7 @@ data class NotificationEvent(
         return try {
             XContentHelper.toXContent(this, XContentType.JSON, EMPTY_PARAMS, true).utf8ToString()
         } catch (e: IOException) {
-            throw OpenSearchException(e)
+            super.toString() + " threw " + e.toString()
         }
     }
 }
--- a/src/main/kotlin/org/opensearch/commons/notifications/model/NotificationEvent.kt
+++ b/src/main/kotlin/org/opensearch/commons/notifications/model/NotificationEvent.kt
@@ -4,13 +4,17 @@
  */
 package org.opensearch.commons.notifications.model
 
+import org.opensearch.OpenSearchException
 import org.opensearch.common.io.stream.StreamInput
 import org.opensearch.common.io.stream.StreamOutput
 import org.opensearch.common.io.stream.Writeable
 import org.opensearch.common.xcontent.ToXContent
+import org.opensearch.common.xcontent.ToXContent.EMPTY_PARAMS
 import org.opensearch.common.xcontent.XContentBuilder
+import org.opensearch.common.xcontent.XContentHelper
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/opensearch/commons/notifications/model/NotificationEvent.kt
```diff
@@ -102,4 +105,13 @@ data class NotificationEvent(
             .field(STATUS_LIST_TAG, statusList)
             .endObject()
     }
+
+    // Overriding toString so consuming plugins can log/output this from the  sendNotification response if needed
+    override fun toString(): String {
+        return try {
+            XContentHelper.toXContent(this, XContentType.JSON, EMPTY_PARAMS, true).utf8ToString()
+        } catch (e: IOException) {
+            super.toString() + " threw " + e.toString()
```
rishabhmaurya, line 114: Are we publishing this information outside of the system?

rishabhmaurya:
Can you serialize and deserialize logic test for `LegacyPublishNotificartionRequest` with destination type `LEGACY_EMAIL` - https://github.com/qreshi/opensearch-common-utils/blob/email-via-passthrough/src/test/kotlin/org/opensearch/commons/notifications/action/LegacyPublishNotificationRequestTests.kt

src/main/kotlin/org/opensearch/commons/notifications/model/NotificationEvent.kt
qreshi: Alerting is using the response [after publishing a message](https://github.com/opensearch-project/alerting/pull/401/files#diff-ffc48988398120ee598d906b522ee27809ff0712c3eb63a2432b38a4d782fc6dR140) and setting that as the [actionResponseContent in runAction](https://github.com/opensearch-project/alerting/pull/401/files#diff-1afa0c5b81f1bd6459dc8398dd5f24cc989b9fa3447541519a50dc5a50e06286R694).

Since this is the [actionResult](https://github.com/opensearch-project/alerting/pull/401/files#diff-1afa0c5b81f1bd6459dc8398dd5f24cc989b9fa3447541519a50dc5a50e06286R521) and we store it in the Alert it can be published externally via `ctx.alert` in messages.

I had to catch the `IOException` since detekt was complaining that `toString()` should not throw exceptions. If there are concerns here, should I log `super.toString() + " threw " + e.toString()` instead and make the returned string something generic?

rishabhmaurya: but this would suppress the exception? which might be misleading for its consumers.

qreshi: From what I've read, it seems it's not that unusual to suppress the exception here. Conventionally, `toString()` should not be throwing exceptions.

Here is an example of a thread that I found when looking into it: https://stackoverflow.com/a/42011744

This is probably why the linter also failed on it. I can add a log entry as a follow-up change so the exception is logged at least for debugging purposes. If you disagree with this though, feel free to create an issue for it to discuss further.

getsaurabh02: +1 on debug logger here instead

getsaurabh02 (commented):
LGTM overall. Few minor comments to address. Thanks @qreshi

rishabhmaurya (approved):

getsaurabh02 (approved):

PR #222, opened 2022-08-22
Added requestedTenant access info in User model

vamsi-amazon:
Signed-off-by: vamsi-amazon <reddyvam@amazon.com>

### Description
Added requestedTenancy access info in User Model.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/commons/authuser/User.java
+++ b/src/main/java/org/opensearch/commons/authuser/User.java
@@ -40,6 +40,7 @@ final public class User implements Writeable, ToXContent {
     public static final String ROLES_FIELD = "roles";
     public static final String CUSTOM_ATTRIBUTE_NAMES_FIELD = "custom_attribute_names";
     public static final String REQUESTED_TENANT_FIELD = "user_requested_tenant";
+    public static final String REQUESTED_TENANT_ACCESS = "user_requested_tenant_access";
 
     private final String name;
     private final List<String> backendRoles;
@@ -47,6 +48,8 @@ final public class User implements Writeable, ToXContent {
     private final List<String> customAttNames;
     @Nullable
     private final String requestedTenant;
+    @Nullable
+    private final String requestedTenantAccess;
 
     public User() {
         name = "";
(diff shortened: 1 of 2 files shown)


getsaurabh02 (commented):
Which release are planning to release this? Also do we have any backport plans considering this could be a breaking change for the consumer plugins.

vamsi-amazon:
> Which release are planning to release this? Also do we have any backport plans considering this could be a breaking change for the consumer plugins.

I want to backport to all the versions till 1.3. 

I am not sure of all the consumers of the plugin. If the consumers are creating User then this would cause issues if they don't upgrade because of the changes in constructor. 

Should I keep the old constructor as well? I guess this is the safest way so that whoever is using old constructor would not cause any failures.

dblock:
> Should I keep the old constructor as well? I guess this is the safest way so that whoever is using old constructor would not cause any failures.

Yes, that's probably the right thing to do.

src/main/java/org/opensearch/commons/authuser/User.java
```diff
@@ -62,20 +66,38 @@ public User(final String name, final List<String> backendRoles, List<String> rol
...
         this.customAttNames = customAttNames;
         this.requestedTenant = null;
+        this.requestedTenantAccess = null;
+    }
+
+    public User(
+        final String name,
+        final List<String> backendRoles,
+        List<String> roles,
+        List<String> customAttNames,
```
getsaurabh02, line 76: nit: final

getsaurabh02 (changes requested):

getsaurabh02 (approved):

src/main/java/org/opensearch/commons/authuser/User.java
```diff
@@ -40,20 +40,24 @@ final public class User implements Writeable, ToXContent {
...
+    public static final String REQUESTED_TENANT_ACCESS = "user_requested_tenant_access";
 
     private final String name;
     private final List<String> backendRoles;
     private final List<String> roles;
     private final List<String> customAttNames;
     @Nullable
     private final String requestedTenant;
+    @Nullable
+    private final String requestedTenantAccess;
```
peternied, line 52: Instead of treating this as a string can we make it an enum?

PR #281, opened 2022-10-13
Migrate to Apache HttpClient / Core 5.x

### LINKED ISSUE #279: [AUTOCUT] OS Distribution Build Failed for common-utils-3.0.0
***Received Error***: **Error building common-utils, retry with: ./build.sh manifests/3.0.0/opensearch-3.0.0.yml --component common-utils --snapshot**.
                      The distribution build for common-utils has failed.
                      Please see build log at https://build.ci.opensearch.org/job/distribution-build-opensearch/6156/consoleFull

sbcd90:
Signed-off-by: Subhobrata Dey <sbcd90@gmail.com>

### Description
this PR fixes the build failures in `common-utils` due to `upgrade of Apache HttpClient / Core to 5.x` in `opensearch-core`.
https://github.com/opensearch-project/OpenSearch/commit/4833e080cc506237fee5aa90767d6f49d952bec3
 
### Issues Resolved
#279
 
### Check List
- [X] New functionality includes testing.
  - [X] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/commons/authuser/User.java
+++ b/src/main/java/org/opensearch/commons/authuser/User.java
@@ -14,7 +14,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
-import org.apache.http.util.EntityUtils;
+import org.apache.hc.core5.http.ParseException;
+import org.apache.hc.core5.http.io.entity.EntityUtils;
 import org.opensearch.client.Response;
 import org.opensearch.common.Nullable;
 import org.opensearch.common.Strings;
@@ -83,7 +84,7 @@ final public class User implements Writeable, ToXContent {
      * @param response
      * @throws IOException
      */
-    public User(final Response response) throws IOException {
+    public User(final Response response) throws IOException, ParseException {
         this(EntityUtils.toString(response.getEntity()));
     }
 
--- a/src/test/kotlin/org/opensearch/commons/alerting/model/ClusterMetricsInputTests.kt
+++ b/src/test/kotlin/org/opensearch/commons/alerting/model/ClusterMetricsInputTests.kt
@@ -91,7 +91,7 @@ class ClusterMetricsInputTests {
         // GIVEN
         path = "/_cluster/health/"
         pathParams = "index1,index2,index3,index4,index5"
-        url = "http://localhost:9200/_cluster/health/index1,index2,index3,index4,index5"
+        url = "http://localhost:9200/_cluster/health/index1%2Cindex2%2Cindex3%2Cindex4%2Cindex5"
 
         // WHEN
         val clusterMetricsInput = ClusterMetricsInput(path, pathParams, url)
@@ -207,7 +207,7 @@ class ClusterMetricsInputTests {
         // GIVEN
         path = "/_cluster/health/"
         pathParams = "index1,index2,index3,index4,index5"
-        val testUrl = "http://localhost:9200/_cluster/health/index1,index2,index3,index4,index5"
+        val testUrl = "http://localhost:9200/_cluster/health/index1%2Cindex2%2Cindex3%2Cindex4%2Cindex5"
         val clusterMetricsInput = ClusterMetricsInput(path, pathParams, url)
 
         // WHEN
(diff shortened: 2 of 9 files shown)


eirsep (dismissed):

src/main/java/org/opensearch/commons/authuser/User.java
```diff
@@ -83,7 +84,7 @@ public User(
      * @param response
      * @throws IOException
      */
-    public User(final Response response) throws IOException {
+    public User(final Response response) throws IOException, ParseException {
```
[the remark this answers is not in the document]
sbcd90: thanks a lot @eirsep for review.
i think we cannot avoid it as it is thrown by the library method itself unless we want to transform it to `IOException` which may be misinterpreted. looks like core changes also throw the error https://github.com/opensearch-project/OpenSearch/blob/4833e080cc506237fee5aa90767d6f49d952bec3/client/rest-high-level/src/test/java/org/opensearch/client/OpenSearchRestHighLevelClientTestCase.java#L329

src/test/kotlin/org/opensearch/commons/alerting/model/ClusterMetricsInputTests.kt
```diff
@@ -437,6 +437,6 @@ class ClusterMetricsInputTests {
         // THEN
         assertEquals(path, clusterMetricsInput.path)
         assertEquals(pathParams, clusterMetricsInput.pathParams)
-        assertEquals(testUrl, clusterMetricsInput.url)
+        assertEquals(testUrl.replace(",", "%2C"), clusterMetricsInput.url)
```
dblock, line 440: I suppose it's fine for a test, but you're URL-encoding/decoding here, so you should be calling url decode.

src/test/kotlin/org/opensearch/commons/alerting/model/ClusterMetricsInputTests.kt
```diff
@@ -91,7 +91,7 @@ class ClusterMetricsInputTests {
         // GIVEN
         path = "/_cluster/health/"
         pathParams = "index1,index2,index3,index4,index5"
-        url = "http://localhost:9200/_cluster/health/index1,index2,index3,index4,index5"
+        url = "http://localhost:9200/_cluster/health/index1%2Cindex2%2Cindex3%2Cindex4%2Cindex5"
```
dblock, line 94: This feels unexpected. Did the API use to take URL-decoded URL and pass it as is, but now requires the arguments to be url-encoded? Either way I think you want to be explicit and call a URL encode function instead of changing all `,` by `%2C`.

dblock (dismissed):
Left a comment on url encoding/decoding in tests. Consider it before merging.

eirsep (approved):

src/test/kotlin/org/opensearch/commons/alerting/model/ClusterMetricsInputTests.kt
sbcd90: @dblock fixed the logic so tests no longer need changes. https://github.com/opensearch-project/common-utils/blob/5a0b10dd4af42abd4099e846328deaeefc2166f5/src/main/kotlin/org/opensearch/commons/alerting/model/ClusterMetricsInput.kt#L208-#L218
Looks like it is a problem with `httpcomponents 5.1.x` library as it has inconsistent behavior across 2 of its constructors.

this encodes the path params
```
val uriBuilder = URIBuilder()
                .setScheme(SUPPORTED_SCHEME)
                .setHost(SUPPORTED_HOST)
                .setPort(SUPPORTED_PORT)
                .setPath(path + pathParams)
```

while this doesnt

```
val uriBuilder = URIBuilder("$SUPPORTED_SCHEME://$SUPPORTED_HOST:$SUPPORTED_PORT$path$pathParams")
```
As this class is used by `alerting` plugin i would like to keep the tests same as any change in behavior may break `alerting` cluster metrics scenario.
Kindly have a look.

src/test/kotlin/org/opensearch/commons/alerting/model/ClusterMetricsInputTests.kt
sbcd90: @dblock fixed the logic so tests no longer need changes. kindly check my comment below.

getsaurabh02 (approved):

PR #283, opened 2022-10-13
Move acknowledge alerts dtos from alerting to common-utils

eirsep:
Signed-off-by: Surya Sashank Nistala <snistala@amazon.com>


eirsep:
https://github.com/opensearch-project/common-utils/pull/289
covered this change

PR #396, opened 2023-04-11
InjectSecurity - inject User object in UserInfo in threadContext

petardz:
### Description
Added user_info injection of User object in InjectSecurity

Pre-requisite for https://github.com/opensearch-project/alerting/pull/852
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/commons/InjectSecurity.java
+++ b/src/main/java/org/opensearch/commons/InjectSecurity.java
@@ -10,12 +10,14 @@ import static org.opensearch.commons.ConfigConstants.OPENSEARCH_SECURITY_INJECTE
 import static org.opensearch.commons.ConfigConstants.OPENSEARCH_SECURITY_USE_INJECTED_USER_FOR_PLUGINS;
 
 import java.util.List;
+import java.util.StringJoiner;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.common.Strings;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.commons.authuser.User;
 
 /**
  * For background jobs usage only. User or Roles injection can be done using transport layer only.
@@ -104,7 +106,7 @@ public class InjectSecurity implements AutoCloseable {
 
     /**
      * Injects user.
-     * @param user
+     * @param user name
      */
     public void injectUser(final String user) {
         if (Strings.isNullOrEmpty(user)) {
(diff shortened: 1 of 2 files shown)


lezzago (dismissed):

src/main/java/org/opensearch/commons/InjectSecurity.java
```diff
@@ -115,8 +117,27 @@ public void injectUser(final String user) {
...
+
+    /**
+     * Injects user object into user info.
+     * @param user
+     */
+    public void injectUserInfo(final User user) {
+        if (user == null) {
+            return;
+        }
+        if (threadContext.getTransient(ConfigConstants.OPENSEARCH_SECURITY_USER_INFO_THREAD_CONTEXT) != null) {
```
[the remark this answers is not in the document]
bowenlan-amzn: Yeah, the error message may not be that useful. 
Can you add in the comment of this method that this should only be used within plugin after stash the context and injecting user permission? I think this is the only use case of this method.

src/main/java/org/opensearch/commons/InjectSecurity.java
```diff
@@ -115,8 +117,27 @@ public void injectUser(final String user) {
...
+            log.error("{}, InjectSecurity - most likely thread context corruption : {}", Thread.currentThread().getName(), id);
+            return;
         }
+        StringJoiner joiner = new StringJoiner("|");
+        joiner.add(user.getName());
+        joiner.add(java.lang.String.join(",", user.getBackendRoles()));
+        joiner.add(java.lang.String.join(",", user.getRoles()));
```
bowenlan-amzn, line 139: Is there already a method to reuse for combining these?

src/main/java/org/opensearch/commons/InjectSecurity.java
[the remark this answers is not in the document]
petardz: @eirsep @bowenlan-amzn 

Added comments, better logging and unit test

bowenlan-amzn (dismissed):
Thanks!

lezzago (approved):

eirsep (approved):

lezzago (approved):

src/main/java/org/opensearch/commons/InjectSecurity.java
cwperks: This is the source of where this ThreadContext header is populated from the security plugin: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L197-L209

This is done in the `PrivilegesEvaluator` which is called from the `SecurityFilter` which is the first action filter that is applied before a transport request is executed on a node. 

`mappedRoles` in this method is the set of roles that are resolved to as part of the roles resolution process which is calculating via a roles mapping like:

```
# mapping for role1
role1
{
  "backend_roles" : [ "starfleet", "captains", "defectors" ],
  "hosts" : [ "*.starfleetintranet.com" ],
  "users" : [ "kirk", "spock" ],
  "and_backend_roles": ["enterprise", "voyager"] // User must have all backend roles in this list to be mapped to role1
}
```

bowenlan-amzn: @cwperks thanks for checking on this. If there's any possible regression in the future, please help track it in some issue.

cwperks: @bowenlan-amzn No regression that I can see. I saw this PR and wanted to get familiar with what it was introducing. I thought I'd leave a fly-by comment for future reference in case anyone wants to know where this threadcontext header originally comes from.

PR #420, opened 2023-05-02
NoOpTrigger

petardz:
### Description
Added NoOpTrigger to be used when constructing monitor error alerts or alerts which are not generated as a result of ran triggers
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/alerting/model/NoOpTrigger.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/model/NoOpTrigger.kt
@@ -1,0 +1,75 @@
+package org.opensearch.commons.alerting.model
+
+import org.opensearch.common.CheckedFunction
+import org.opensearch.common.UUIDs
+import org.opensearch.common.io.stream.StreamInput
+import org.opensearch.common.io.stream.StreamOutput
+import org.opensearch.common.xcontent.XContentParserUtils
+import org.opensearch.commons.alerting.model.action.Action
+import org.opensearch.core.ParseField
+import org.opensearch.core.xcontent.NamedXContentRegistry
+import org.opensearch.core.xcontent.ToXContent
+import org.opensearch.core.xcontent.XContentBuilder
+import org.opensearch.core.xcontent.XContentParser
+import java.io.IOException
+
+data class NoOpTrigger(
+    override val id: String = UUIDs.base64UUID(),
+    override val name: String = "NoOp trigger",
+    override val severity: String = "",
+    override val actions: List<Action> = listOf(),
+) : Trigger {
+
+    @Throws(IOException::class)
+    constructor(sin: StreamInput) : this()
+
+    override fun toXContent(builder: XContentBuilder, params: ToXContent.Params): XContentBuilder {
+        builder.startObject()
+            .startObject(NOOP_TRIGGER_FIELD)
+            .endObject()
+        return builder
+    }
+
+    override fun name(): String {
+        return NOOP_TRIGGER_FIELD
+    }
+
+    fun asTemplateArg(): Map<String, Any> {
+        return mapOf()
+    }
+
+    @Throws(IOException::class)
+    override fun writeTo(out: StreamOutput) {
+    }
+
+    companion object {
+        const val NOOP_TRIGGER_FIELD = "noop_trigger"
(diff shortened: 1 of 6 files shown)


eirsep (dismissed):

eirsep:
plz resolve conflicts

lezzago:
Is the idea to use the NoOpTriggers to create error alerts for all monitor types when the monitor fails?

petardz:
yes

src/main/kotlin/org/opensearch/commons/alerting/model/NoOpTrigger.kt
```diff
@@ -0,0 +1,75 @@
...
+import org.opensearch.common.xcontent.XContentParserUtils
+import org.opensearch.commons.alerting.model.action.Action
+import org.opensearch.core.ParseField
+import org.opensearch.core.xcontent.NamedXContentRegistry
+import org.opensearch.core.xcontent.ToXContent
+import org.opensearch.core.xcontent.XContentBuilder
+import org.opensearch.core.xcontent.XContentParser
+import java.io.IOException
+
+data class NoOpTrigger(
```
lezzago, line 16: Is the only difference from a normal trigger is that it has no condition, so its technically always true?

petardz: Yea it has id/name so that Alert's fields can be populated. Also, creating monitor with this trigger is not permitted.(require statement in init of monitor model)

petardz:
> > yes
> 
> How exactly will the NoOptrigger be used in the monitor executions? Also why is this needed for the error alerts? Can't we have a function in Alerting to generate error alerts from monitor failures and have that shared amongst all the monitor runners? Something similar to that is already in place.
> 
> Also would the idea be for users to make monitors with these noop triggers?

After discussion with @getsaurabh02 conclusion was that it is safest to keep constraint of mandatory trigger parameter in Alert constructors, hence the NoOpTrigger implementation.

You can't make monitors with this type of trigger. I added this constraint inside init block of Monitor class.

src/test/kotlin/org/opensearch/commons/alerting/model/XContentTests.kt
```diff
@@ -357,6 +358,16 @@ class XContentTests {
         assertEquals("Round tripping alert doesn't work", alert, parsedAlert)
     }
 
+    @Test
```
lezzago, line 361: Lets add xcontent tests for NoOp triggers.

src/main/kotlin/org/opensearch/commons/alerting/model/NoOpTrigger.kt
```diff
@@ -51,19 +54,17 @@ data class NoOpTrigger(
 
         @JvmStatic @Throws(IOException::class)
         fun parseInner(xcp: XContentParser): NoOpTrigger {
-            if (xcp.currentToken() != XContentParser.Token.START_OBJECT && xcp.currentToken() != XContentParser.Token.FIELD_NAME) {
-                XContentParserUtils.throwUnknownToken(xcp.currentToken(), xcp.tokenLocation)
-            }
-
-            // If the parser began on START_OBJECT, move to the next token so that the while loop enters on
-            // the fieldName (or END_OBJECT if it's empty).
+            var id = ""
```
lezzago, line 57: Lets generate an id in case we want to have this used by monitors. Something llike:
```
var id = UUIDs.base64UUID()
```

src/test/kotlin/org/opensearch/commons/alerting/model/XContentTests.kt
```diff
@@ -180,6 +180,16 @@ class XContentTests {
...
+    fun `test no-op trigger parsing`() {
+        val trigger = NoOpTrigger()
+
+        val triggerString = trigger.toXContent(builder(), ToXContent.EMPTY_PARAMS).string()
+        val parsedTrigger = Trigger.parse(parser(triggerString))
+
+        Assertions.assertEquals(trigger, parsedTrigger, "Round tripping BucketLevelTrigger doesn't work")
```
lezzago, line 190: Fix the text of the error message to mention no-op trigger instead

eirsep (approved):

PR #380, opened 2023-03-13
Added workflow as a composite monitor

stevanbz:
### Description
Added workflow
 
### Issues Resolved
[List any issues this PR will resolve]
 
### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #439, opened 2023-05-24
Adds workflows model and workflow actions for Alerting Plugin 

eirsep:
PRs merged into feature branch feature/composite-monitors:
Added Alerting workflow model (https://github.com/opensearch-project/common-utils/pull/380)
Renamed chained monitor findings (https://github.com/opensearch-project/common-utils/pull/390)
Imports fixed (https://github.com/opensearch-project/common-utils/pull/394)
Added validation on IndexWorkflowRequest class (https://github.com/opensearch-project/common-utils/pull/405)
Extended deleted workflow response with list of monitor ids failed to delete (https://github.com/opensearch-project/common-utils/pull/416)

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/alerting/action/IndexWorkflowRequest.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/action/IndexWorkflowRequest.kt
@@ -1,0 +1,137 @@
+package org.opensearch.commons.alerting.action
+
+import org.opensearch.action.ActionRequest
+import org.opensearch.action.ActionRequestValidationException
+import org.opensearch.action.ValidateActions
+import org.opensearch.action.support.WriteRequest
+import org.opensearch.common.io.stream.StreamInput
+import org.opensearch.common.io.stream.StreamOutput
+import org.opensearch.commons.alerting.model.CompositeInput
+import org.opensearch.commons.alerting.model.Workflow
+import org.opensearch.rest.RestRequest
+import java.io.IOException
+import java.util.stream.Collectors
+
+class IndexWorkflowRequest : ActionRequest {
+    val workflowId: String
+    val seqNo: Long
+    val primaryTerm: Long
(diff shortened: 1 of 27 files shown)


lezzago (dismissed):

src/main/kotlin/org/opensearch/commons/alerting/action/IndexWorkflowRequest.kt
```diff
@@ -0,0 +1,137 @@
...
+        var validationException: ActionRequestValidationException? = null
+
+        if (workflow.inputs.isEmpty()) {
+            validationException = ValidateActions.addValidationError(
+                "Input list can not be empty.", validationException
+            )
+            return validationException
+        }
+        if (workflow.inputs.size > 1) {
+            validationException = ValidateActions.addValidationError(
```
getsaurabh02, line 65: Looks like some of these flows are not covered by the unit test. Consider looking into ``Codecov`` comments.

getsaurabh02 (commented):
I see three different names being used here in the PR - Workflows, ChainedMonitor and Composite. Wondering if we need to arrive at the consistency in naming, especially when UX and Rest APIs should be using Composite Monitors as the name.

src/main/kotlin/org/opensearch/commons/alerting/action/IndexWorkflowRequest.kt
eirsep: > I see three different names being used here in the PR - Workflows, ChainedMonitor and Composite. Wondering if we need to arrive at the consistency in naming, especially when UX and Rest APIs should be using Composite Monitors as the name.

**Workflow** refers to the the new type of scheduled job.
**Composite** is the input type - it accepts a sequence of monitors to be executed as input.
**Chained Monitor** refers to chained monitor findings i.e. findings of one monitor are used as the source data for the next monitor which is 'chained' to it.

lezzago (approved):

PR #458, opened 2023-06-21
add ui metadata in workflow object

eirsep:
add ui metadata in worklfow object similar to ui metadata object present in monitor. this is required for dashboards to store metadata information


lezzago:
There has been talks about going away from this logic since if someone creates the monitor/workflow through the api, it can never be converted to the visual mode even if the visual mode would support that configuration.

PR #463, opened 2023-06-21
Add workflowId field in alert model

eirsep:
### Description
Add workflowId field in alert model.
This field would be populated when alerts are created by monitors executing within a workflow. Else field would be an empty string

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
@@ -22,6 +22,7 @@ data class Alert(
     val version: Long = NO_VERSION,
     val schemaVersion: Int = NO_SCHEMA_VERSION,
     val monitorId: String,
+    val workflowId: String,
     val monitorName: String,
     val monitorVersion: Long,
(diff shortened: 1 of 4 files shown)


sbcd90 (approved):

src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
```diff
@@ -22,6 +22,7 @@ data class Alert(
     val version: Long = NO_VERSION,
     val schemaVersion: Int = NO_SCHEMA_VERSION,
     val monitorId: String,
+    val workflowId: String,
```
lezzago, line 25: Shouldn't this be optional? Normal monitor executions will fail when creating an alert since workflows are not there. Also the vice versa might be an issue since the chained alert won't have a monitor id.

eirsep: monitor id would be null or empty for chained alerts
workflow Id would be null or empty for non-workflow alert

in the constructors we handle each case and set "" to handle the empty case.

lezzago (approved):

PR #513, opened 2023-08-30
[Breaking Change] Corrects the security config constant name which uses trusted cas

DarshitChanpura:
### Description
SecureRestClientBuilder currently is using node-certificate constant `pemcert_filepath` to build trust store. This is incorrect. It should be using `pemtrustedcas_filepath` to build a trust-store with root CAs
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/commons/ConfigConstants.java
+++ b/src/main/java/org/opensearch/commons/ConfigConstants.java
@@ -21,7 +21,7 @@ public class ConfigConstants {
     public static final String AUTHORIZATION = "Authorization";
 
     // These reside in security plugin.
-    public static final String OPENSEARCH_SECURITY_SSL_HTTP_PEMCERT_FILEPATH = "plugins.security.ssl.http.pemcert_filepath";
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/commons/ConfigConstants.java
```diff
@@ -21,7 +21,7 @@ public class ConfigConstants {
     public static final String AUTHORIZATION = "Authorization";
 
     // These reside in security plugin.
-    public static final String OPENSEARCH_SECURITY_SSL_HTTP_PEMCERT_FILEPATH = "plugins.security.ssl.http.pemcert_filepath";
+    public static final String OPENSEARCH_SECURITY_SSL_HTTP_PEMTRUSTEDCAS_FILEPATH = "plugins.security.ssl.http.pemtrustedcas_filepath";
```
lezzago, line 24: This is going to break a few plugins with this name change. If we go ahead with this, issues to each of the plugins should be made to indicate this change.

Affected plugins: https://github.com/search?q=org%3Aopensearch-project%20%22OPENSEARCH_SECURITY_SSL_HTTP_PEMCERT_FILEPATH%22&type=code

lezzago: Also is it necessary to have a variable name change here?

PR #514, opened 2023-08-30
support list of monitor ids in Chained Monitor Findings

eirsep:
Adds support for passing a list monitor ids in chained monitor findings. This helps filter data of a monitor in workflow based on findings of multiple monitors instead of a single monitor.

If workflow sequence looks like: 
[m1] -> [m2] -> [m3 (chained findings of m1, m2)]
this means m3's data input would be the related doc ids from findings of m1 and m2.

for backward compatibility reasons, the monitorId field would be given preference over the new list type field, monitorIds, if both are provided.

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/alerting/action/IndexWorkflowRequest.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/action/IndexWorkflowRequest.kt
@@ -30,7 +30,7 @@ class IndexWorkflowRequest : ActionRequest {
         refreshPolicy: WriteRequest.RefreshPolicy,
         method: RestRequest.Method,
         workflow: Workflow,
-        rbacRoles: List<String>? = null
+        rbacRoles: List<String>? = null,
     ) : super() {
         this.workflowId = workflowId
         this.seqNo = seqNo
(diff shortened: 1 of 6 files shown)


lezzago (approved):

src/main/kotlin/org/opensearch/commons/alerting/action/IndexWorkflowRequest.kt
```diff
@@ -30,7 +30,7 @@ class IndexWorkflowRequest : ActionRequest {
         refreshPolicy: WriteRequest.RefreshPolicy,
         method: RestRequest.Method,
         workflow: Workflow,
-        rbacRoles: List<String>? = null
+        rbacRoles: List<String>? = null,
```
sbcd90, line 33: is the `,` required?

src/main/kotlin/org/opensearch/commons/alerting/model/ChainedMonitorFindings.kt
```diff
@@ -9,63 +9,89 @@ import org.opensearch.core.xcontent.XContentBuilder
...
+        } else {
+            monitorIds.forEach { validateId(it) }
+        }
     }
 
     @Throws(IOException::class)
     constructor(sin: StreamInput) : this(
-        sin.readString(), // monitorId
+        sin.readOptionalString(), // monitorId
+        Collections.unmodifiableList(sin.readStringList())
```
sbcd90, line 37: can we use kotlin `toList` here?

src/main/kotlin/org/opensearch/commons/alerting/model/ChainedMonitorFindings.kt
```diff
@@ -9,63 +9,89 @@ import org.opensearch.core.xcontent.XContentBuilder
...
     )
 
+    @Suppress("UNCHECKED_CAST")
     fun asTemplateArg(): Map<String, Any> {
         return mapOf(
             MONITOR_ID_FIELD to monitorId,
-        )
+            MONITOR_IDS_FIELD to monitorIds
```
sbcd90, line 44: is it possible we merge these 2 fields to 1? it looks little complex that we have both
```
MONITOR_ID_FIELD to monitorId,
MONITOR_IDS_FIELD to monitorIds```

eirsep: we need to have for backward compatibility

src/main/kotlin/org/opensearch/commons/alerting/action/IndexWorkflowRequest.kt
eirsep: done

sbcd90 (approved):

lezzago (approved):

PR #547, opened 2023-10-09
Add User.isAdminDn to User class

cwperks:
### Description

Adds a static helper method on the User class to determine if the user is in the security plugin's `plugins.security.authcz.admin_dn` list.
 
### Issues Resolved

https://github.com/opensearch-project/security/issues/3413
 
### Check List
- [X] New functionality includes testing.
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/commons/authuser/User.java
+++ b/src/main/java/org/opensearch/commons/authuser/User.java
@@ -10,6 +10,7 @@ import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedTok
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -19,8 +20,10 @@ import org.apache.hc.core5.http.io.entity.EntityUtils;
 import org.opensearch.client.Response;
 import org.opensearch.common.Nullable;
 import org.opensearch.common.inject.internal.ToStringBuilder;
+import org.opensearch.common.settings.Settings;
 import org.opensearch.common.xcontent.XContentHelper;
 import org.opensearch.common.xcontent.json.JsonXContent;
+import org.opensearch.commons.ConfigConstants;
 import org.opensearch.core.common.Strings;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
@@ -249,5 +252,10 @@ final public class User implements Writeable, ToXContent {
     @Nullable
     public String getRequestedTenant() {
         return requestedTenant;
+    }
+
+    public static boolean isSuperUser(User user, Settings settings) {
+        List<String> adminDns = settings.getAsList(ConfigConstants.OPENSEARCH_SECURITY_AUTHCZ_ADMIN_DN, Collections.emptyList());
+        return adminDns.contains(user.getName());
     }
 }
--- a/src/main/java/org/opensearch/commons/authuser/User.java
+++ b/src/main/java/org/opensearch/commons/authuser/User.java
@@ -255,6 +255,9 @@ final public class User implements Writeable, ToXContent {
     }
 
     public static boolean isSuperUser(User user, Settings settings) {
+        if (user == null || settings == null) {
+            return false;
+        }
         List<String> adminDns = settings.getAsList(ConfigConstants.OPENSEARCH_SECURITY_AUTHCZ_ADMIN_DN, Collections.emptyList());
         return adminDns.contains(user.getName());
     }
--- a/src/main/java/org/opensearch/commons/authuser/User.java
+++ b/src/main/java/org/opensearch/commons/authuser/User.java
@@ -254,11 +254,11 @@ final public class User implements Writeable, ToXContent {
         return requestedTenant;
     }
 
-    public static boolean isSuperUser(User user, Settings settings) {
-        if (user == null || settings == null) {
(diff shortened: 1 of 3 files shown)


cwperks:
@dhrubo-os @ylwu-amzn Let me know if this helps. There should be some testing performed with the security plugin installed, but this should let you determine if the current user is in the `admin_dn` list (i.e. user is a super user)

src/main/java/org/opensearch/commons/authuser/User.java
```diff
@@ -250,4 +253,9 @@ public List<String> getCustomAttNames() {
     public String getRequestedTenant() {
         return requestedTenant;
     }
+
+    public static boolean isSuperUser(User user, Settings settings) {
+        List<String> adminDns = settings.getAsList(ConfigConstants.OPENSEARCH_SECURITY_AUTHCZ_ADMIN_DN, Collections.emptyList());
+        return adminDns.contains(user.getName());
```
dhrubo-os, line 259: I think we might need some backward compatibility? What happens if user is null? then it will throw exception in line 259. And I'm not a security person, so I don't know how to handle this situation. Just wanted to call out here.

cwperks: Added null check

lezzago: Usually when the user is super admin, the user object is null. [Reference](https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt#L227)

cwperks: Please see details [here](https://github.com/opensearch-project/security/issues/3413#issuecomment-1749928762). 

There are 1 cases of superadmin:

1. Plugin stashed the threadcontext (user == null)
2. User connects to cluster using admin certificate <- scope of this PR

cwperks: Capturing discussion on this PR:

1. User is null means 1 of 2 things - 1) Security is not enabled or 2) plugin is operating in a stashContext block
2. User is non-null then - 1) User is a regular user (including anonymous) or 2) User is a superuser (meaning request comes from admin cert)

This PR pertains to 2) above and also relies on:

1. When security is enabled a user will always be present in the threadcontext
2. The plugin dev will ensure they are not making any calls to read the currently authenticated user from within a stashContext block

dhrubo-os:
Do we need to mark this PR as draft? Are we planning to merge this?

dhrubo-os:
> > Do we need to mark this PR as draft? Are we planning to merge this?
> 
> Marked as ready for review. @dhrubo-os if there's a companion ml-commons PR please share to show how this change will be utilized.

I'll share within next week.

dhrubo-os:
> > > Do we need to mark this PR as draft? Are we planning to merge this?
> > 
> > 
> > Marked as ready for review. @dhrubo-os if there's a companion ml-commons PR please share to show how this change will be utilized.
> 
> I'll share within next week.

Just Raised a [draft PR](https://github.com/opensearch-project/ml-commons/pull/1564/files) to show the example

We are going to use this function [similar like this way](https://github.com/opensearch-project/ml-commons/blob/8ef6cd2e20f3d0d2c1b60cfb2c9b8f814bf55a08/plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java#L142)

Currently for the easy development we added [this method](https://github.com/opensearch-project/ml-commons/blob/8ef6cd2e20f3d0d2c1b60cfb2c9b8f814bf55a08/plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java#L208C5-L226C6) in our ml-commons plugin end. But eventually we want to use [this method](https://github.com/opensearch-project/ml-commons/blob/8ef6cd2e20f3d0d2c1b60cfb2c9b8f814bf55a08/plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java#L248-L255)


Please let me know if you have any question or concern. Thanks.

cwperks:
Hey @dhrubo-os, thank you for providing the linked PRs for example usage. If I'm reading the PR correctly it means that if a regular user creates a model its not hidden, but if a superuser creates a model then its hidden?

Have you considered separate transport actions so that those actions are authorized separately? i.e.:

1. TransportRegisterModelAction
2. TransportRegisterHiddenModelAction

If you have a separate transport action then you can permission users to perform the register hidden model action through a separate "admin-level" role. 

Roles for ml-commons plugin: https://github.com/opensearch-project/security/blob/main/config/roles.yml#L273-L295

While there are already instances in plugins of plugins having to implement their own authz models based on user's backend roles, this will be yet another instance of something that should be the responsibility of the security plugin that is implemented in a plugin.

src/main/java/org/opensearch/commons/authuser/User.java
```diff
@@ -250,4 +253,12 @@ public List<String> getCustomAttNames() {
     public String getRequestedTenant() {
         return requestedTenant;
     }
+
+    public boolean isSuperUser(Settings settings) {
```
shikharj05, line 257: nit: should this be named ``isAdminUser`` or ``isAdminDn``

cwperks: Changed to `isAdminDn`. admin user may be confusing since there could be different levels of admin. AdminDn implies the user connecting with admin certificate.

DarshitChanpura (approved):

lezzago (approved):

PR #605, opened 2024-03-05
Enhance per bucket, and per document monitor notification message ctx.

AWSHurneyt:
### Description
1. Added support for printing sample documents for each bucket in bucket level monitor notification messages. 
2. Added support for printing query/rule info in doc level monitor notification messages.

Accomplished this by adding the `AlertContext` data model, which we can use to store data we only want to expose via the notification message `ctx`. Storing sample documents, or query/rule lists in the alert document would bloat it; this data model allows us to separate the content of the `ctx` from the content of the `Alert` documents.
 
### Issues Resolved
https://github.com/opensearch-project/alerting/issues/1300
https://github.com/opensearch-project/alerting/issues/1396
https://github.com/opensearch-project/alerting/issues/1401
 
### Check List
  - [x] All tests pass


src/main/kotlin/org/opensearch/commons/alerting/model/AlertContext.kt
```diff
@@ -0,0 +1,74 @@
...
+            ASSOCIATED_QUERIES_FIELD to queriesContext,
+            SAMPLE_DOCS_FIELD to sampleDocs
+        )
+
+        // Get the alert template args
+        val templateArgs = super.asTemplateArg().toMutableMap()
+
+        // Add the non-null custom context fields to the alert templateArgs.
+        customContextFields.forEach { (key, value) ->
+            if (value !== null) templateArgs[key] = value
```
riysaxen-amzn, line 65: Minor: 

To ensures null values are not added to the templateArgs

```
customContextFields.forEach { (key, value) ->
    value?.let { templateArgs[key] = it }
}

```

src/main/kotlin/org/opensearch/commons/alerting/model/AlertContext.kt
```diff
@@ -0,0 +1,74 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.commons.alerting.model
+
+/**
+ * This model is a wrapper for [Alert] that should only be used to create a more
+ * informative alert object to enrich mustache template notification messages.
+ */
+data class AlertContext(
```
riysaxen-amzn, line 12: What should be the relationship between Alert and AlertContext, a composition or inheritance?

AWSHurneyt: In this case, I decided to go with inheritance. An AlertContext is an Alert that contains more information for the purposes of enriching the notification messages. We already return the entire alert document in the notification message ctx; this just allows us to expand upon the information that's already there. If we were not concerned with the size of the alert documents, it would make sense for all of this new info to be within the alert itself.

src/main/kotlin/org/opensearch/commons/alerting/model/AlertContext.kt
AWSHurneyt: I'm adding some more unit tests. I'll revise the PR shortly, and include this change.

riysaxen-amzn (approved):

src/main/kotlin/org/opensearch/commons/alerting/model/AlertContext.kt
```diff
@@ -0,0 +1,74 @@
...
+    val alert: Alert,
+    val associatedQueries: List<DocLevelQuery>? = null,
+    val sampleDocs: List<Map<String, Any?>>? = null
+) : Alert(
```
engechas, line 16: Non-blocking

Inheritance seems like overkill here. The context could just have an associated alert as a member variable rather than extending an alert. 

Inheritance usually means that A is a B. In this context, I would argue an AlertContext is not an Alert. But I understand that using inheritance here limits the scope of the changes since AlertContext still evaluates to an Alert when typed variables are used.

engechas (approved):

eirsep (changes requested):

src/main/kotlin/org/opensearch/commons/alerting/model/AlertContext.kt
AWSHurneyt: Yeah, I went with this approach of extending the Alert class because I essentially wanted to build upon the functionality of the Alert data model to make a specialized "Alert" that could be used in instances when we want additional information (in this case, when we want to expose more info in the notification message). I'll revise the PRs to make AlertContext a separate class.

src/main/kotlin/org/opensearch/commons/alerting/model/AlertContext.kt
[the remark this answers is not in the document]
AWSHurneyt: I'll revise the PRs based on this feedback.
https://github.com/opensearch-project/common-utils/pull/605#discussion_r1520329777

src/main/kotlin/org/opensearch/commons/alerting/model/AggregationResultBucket.kt
```diff
@@ -43,6 +43,14 @@ data class AggregationResultBucket(
         return builder
     }
 
+    fun asTemplateArg(): Map<String, Any?> {
```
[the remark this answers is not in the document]
AWSHurneyt: The bucket level trigger `asTemplateArg()` functions returns the condition
https://github.com/opensearch-project/common-utils/blob/main/src/main/kotlin/org/opensearch/commons/alerting/model/BucketLevelTrigger.kt#L73

engechas (approved):

src/main/kotlin/org/opensearch/commons/alerting/model/AlertContext.kt
```diff
@@ -0,0 +1,46 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.commons.alerting.model
+
+/**
+ * This model is a wrapper for [Alert] that should only be used to create a more
+ * informative alert object to enrich mustache template notification messages.
+ */
+data class AlertContext(
```
eirsep, line 12: can this class be written in Alerting itself. seems like a construct not used by any other plugin?

src/test/kotlin/org/opensearch/commons/alerting/model/AlertContextTests.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.opensearch.commons.alerting.randomAlertContext
+
+class AlertContextTests {
+    private var alertContext: AlertContext = randomAlertContext()
+
+    @BeforeEach
+    fun generateRandomData() {
+        alertContext = randomAlertContext()
```
eirsep, line 18: nit: is this required? we seem to be doing this initialization of the `var` itself?

src/test/kotlin/org/opensearch/commons/alerting/TestHelpers.kt
```diff
@@ -601,3 +602,22 @@ fun randomFinding(
         timestamp = timestamp
     )
 }
+
+fun randomAlertContext(
+    alert: Alert = randomAlert(),
+    associatedQueries: List<DocLevelQuery>? = (-1..2).random().takeIf { it != -1 }?.let {
```
eirsep, line 608: can we add 2 or more queries to this alert?

eirsep: can we add tests verifying template-args for 0 queries and 2 queries

src/main/kotlin/org/opensearch/commons/alerting/model/AlertContext.kt
AWSHurneyt: Good point. I had initially added it in common utils because I was inheriting the Alert data model. I'll move it, and close this PR.

src/test/kotlin/org/opensearch/commons/alerting/model/AlertContextTests.kt
AWSHurneyt: Removed.

AWSHurneyt:
Closing this issue as I've moved these changes to the alerting plugin in PR https://github.com/opensearch-project/alerting/pull/1450 based on this feedback https://github.com/opensearch-project/common-utils/pull/605#discussion_r1520545127

PR #667, opened 2024-06-05
Adding replication (CCR) plugin interface and classes to common-utils

aggarwalShivani:
### Description
**Background:**

- This change is required for a new feature being added in ism plugin -  [unfollow-action #726](https://github.com/opensearch-project/index-management/issues/726) - integrating ccr and ism plugins.
- For this feature, **ism plugin needs to invoke stop-replication action from ccr.** As both ccr and ism need to invoke some common code, the **common code can be moved to common-utils**, as done previously for other plugins too. For ex. [notifications plugin](https://github.com/opensearch-project/common-utils/pull/31/files)
- As sharing of libraries also leads to a type-cast and class-loading issue - previously seen with [notification plugin](https://github.com/opensearch-project/notifications/pull/223) - the request object needs to be of a higher-level class from opensearch-core like **ActionRequest** and later be recreated into required type i.e. **StopIndexReplicationRequest**.

**Proposed Solution:**
1. Common-utils: Move common code of stop-replication from ccr project to common-utils.
2. CCR: Modify ccr plugin to consume classes from common-utils. Also, create a new TransportAction that transforms the request into required type and invokes TransportStopIndexReplicationAction. 
3. ISM: Add new action in ism
(FYI - This is an alternative approach for this feature. I had also raised a [draft PR](https://github.com/opensearch-project/OpenSearch/pull/13615) with a different approach, but we didnt go ahead with it as it required changes in opensearch-core as well.)

**Change description:**
Code for StopIndexReplicationRequest and StopIndexReplicationAction are moved to common-utils. Added UTs for both the new classes added.
This PR caters to point 1 of the proposed solution.

### Issues Resolved
Related Issues
[unfollow-action #726](https://github.com/opensearch-project/index-management/issues/726)

### Check List
  - [x] All tests pass
- ~[ ] New functionality has been documented.~
  - ~[ ] New functionality has javadoc added~

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/replication/ReplicationPluginInterface.kt
+++ b/src/main/kotlin/org/opensearch/commons/replication/ReplicationPluginInterface.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.commons.replication
+
+import org.opensearch.action.support.master.AcknowledgedResponse
+import org.opensearch.client.node.NodeClient
+import org.opensearch.commons.replication.action.StopIndexReplicationRequest
+import org.opensearch.commons.replication.action.ReplicationActions.STOP_REPLICATION_ACTION_TYPE
+import org.opensearch.commons.utils.recreateObject
+import org.opensearch.core.action.ActionListener
+import org.opensearch.core.action.ActionResponse
+import org.opensearch.core.common.io.stream.Writeable
+
+
+/**
+ * All the transport action plugin interfaces for the cross-cluster-replication plugin.
+ */
+object ReplicationPluginInterface {
+
+    /**
+     * Stop replication.
+     * @param client Node client for making transport action
+     * @param request The request object
+     * @param listener The listener for getting response
+     */
+
+    fun stopReplication(
+        client: NodeClient,
+        request: StopIndexReplicationRequest,
+        listener: ActionListener<AcknowledgedResponse>
+    ) {
+        return client.execute(
+            STOP_REPLICATION_ACTION_TYPE,
+            request,
+            wrapActionListener(listener) { response ->
+                recreateObject(response) {
+                    AcknowledgedResponse(it)
+                }
+            }
+        )
+    }
+
+    /**
+     * Wrap action listener on concrete response class by a new created one on ActionResponse.
+     * This is required because the response may be loaded by different classloader across plugins.
+     * The onResponse(ActionResponse) avoids type cast exception and give a chance to recreate
+     * the response object.
+     */
+    @Suppress("UNCHECKED_CAST")
+    private fun <Response : AcknowledgedResponse> wrapActionListener(
+        listener: ActionListener<Response>,
+        recreate: (Writeable) -> Response
+    ): ActionListener<Response> {
+        return object : ActionListener<ActionResponse> {
+            override fun onResponse(response: ActionResponse) {
+                val recreated = response as? Response ?: recreate(response)
+                listener.onResponse(recreated)
+            }
+
+            override fun onFailure(exception: java.lang.Exception) {
+                listener.onFailure(exception)
+            }
+        } as ActionListener<Response>
+    }
+}
--- a/src/main/kotlin/org/opensearch/commons/replication/action/ReplicationActions.kt
+++ b/src/main/kotlin/org/opensearch/commons/replication/action/ReplicationActions.kt
@@ -1,0 +1,27 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.commons.replication.action
+
+import org.opensearch.action.ActionType
+import org.opensearch.action.support.master.AcknowledgedResponse
+
+/**
+ * All the transport action information for the Replication plugin
+ */
+object ReplicationActions {
+
+    /**
+     * Stop replication. Internal only - Inter plugin communication.
+     */
+    const val STOP_REPLICATION_NAME = "indices:admin/plugins/replication/index/stop"
+    const val STOP_REPLICATION_BASE_ACTION_NAME = "indices:admin/plugins/replication/index/unfollow"
+
+    /**
+     * Stop replication transport action type. Internal only - Inter plugin communication.
+     */
+    val STOP_REPLICATION_ACTION_TYPE =
+        ActionType(STOP_REPLICATION_NAME, ::AcknowledgedResponse)
+
+}
--- a/src/test/kotlin/org/opensearch/commons/replication/ReplicationPluginInterfaceTests.kt
+++ b/src/test/kotlin/org/opensearch/commons/replication/ReplicationPluginInterfaceTests.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.commons.replication
+
+import com.nhaarman.mockitokotlin2.whenever
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.mockito.Answers
+import org.mockito.ArgumentMatchers
+import org.mockito.Mock
+import org.mockito.Mockito
+import org.mockito.junit.jupiter.MockitoExtension
+import org.opensearch.action.ActionType
+import org.opensearch.action.support.master.AcknowledgedResponse
+import org.opensearch.client.node.NodeClient
+import org.opensearch.commons.replication.action.StopIndexReplicationRequest
+import org.opensearch.core.action.ActionListener
+
+@Suppress("UNCHECKED_CAST")
+@ExtendWith(MockitoExtension::class)
+internal class ReplicationPluginInterfaceTests {
+
+    @Mock(answer = Answers.RETURNS_DEEP_STUBS)
+    private lateinit var client: NodeClient
+    @Test
+    fun stopReplication() {
+        val request = Mockito.mock(StopIndexReplicationRequest::class.java)
+        val response = AcknowledgedResponse(true)
+        val listener: ActionListener<AcknowledgedResponse> =
+            Mockito.mock(ActionListener::class.java) as ActionListener<AcknowledgedResponse>
+
+        Mockito.doAnswer {
+            (it.getArgument(2) as ActionListener<AcknowledgedResponse>)
+                .onResponse(response)
+        }.whenever(client).execute(Mockito.any(ActionType::class.java), Mockito.any(), Mockito.any())
+
+        ReplicationPluginInterface.stopReplication(client, request, listener)
+        Mockito.verify(listener, Mockito.times(1)).onResponse(ArgumentMatchers.eq(response))
+    }
+}
--- a/src/main/kotlin/org/opensearch/commons/replication/action/StopIndexReplicationRequest.kt
+++ b/src/main/kotlin/org/opensearch/commons/replication/action/StopIndexReplicationRequest.kt
@@ -1,0 +1,62 @@
+package org.opensearch.commons.replication.action
+
+import org.opensearch.action.ActionRequestValidationException
+import org.apache.logging.log4j.LogManager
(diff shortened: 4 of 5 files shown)


src/main/kotlin/org/opensearch/commons/replication/action/ReplicationActions.kt
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Stop replication. Internal only - Inter plugin communication.
+     */
+    const val STOP_REPLICATION_NAME = "indices:admin/plugins/replication/index/stop"
+    const val STOP_REPLICATION_BASE_ACTION_NAME = "indices:admin/plugins/replication/index/unfollow"
+
+    /**
+     * Stop replication transport action type. Internal only - Inter plugin communication.
+     */
+    val STOP_REPLICATION_ACTION_TYPE =
```
bowenlan-amzn, line 24: Can we define different ActionType for base and stop replication action? So from plugin, we would invoke the TransportAction using the ActionType of base action and requires both 2 permissions. While for the REST API, everything is kept like before.

aggarwalShivani: > Can we define different ActionType for base and stop replication action? So from plugin, we would invoke the TransportAction using the ActionType of base action and requires both 2 permissions. While for the REST API, everything is kept like before.

Yes, thanks for the suggestion @bowenlan-amzn. I've done the same now. 
Added UNFOLLOW_REPLICATION_ACTION_TYPE for plugin, while STOP_REPLICATION_ACTION_TYPE would be used for REST API.

src/main/kotlin/org/opensearch/commons/replication/ReplicationPluginInterface.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * Wrap action listener on concrete response class by a new created one on ActionResponse.
+     * This is required because the response may be loaded by different classloader across plugins.
+     * The onResponse(ActionResponse) avoids type cast exception and give a chance to recreate
+     * the response object.
+     */
+    @Suppress("UNCHECKED_CAST")
+    private fun <Response : AcknowledgedResponse> wrapActionListener(
```
bowenlan-amzn, line 51: From the existing code in AlertingPluginInterface, I see we are doing <Response: BaseResponse>, does that have problem here? can we also use BaseResponse, instead of AcknowledgedResponse.

src/test/kotlin/org/opensearch/commons/replication/ReplicationPluginInterfaceTests.kt
```diff
@@ -0,0 +1,44 @@
...
+        val listener: ActionListener<AcknowledgedResponse> =
+            Mockito.mock(ActionListener::class.java) as ActionListener<AcknowledgedResponse>
+
+        Mockito.doAnswer {
+            (it.getArgument(2) as ActionListener<AcknowledgedResponse>)
+                .onResponse(response)
+        }.whenever(client).execute(Mockito.any(ActionType::class.java), Mockito.any(), Mockito.any())
+
+        ReplicationPluginInterface.stopReplication(client, request, listener)
+        Mockito.verify(listener, Mockito.times(1)).onResponse(ArgumentMatchers.eq(response))
```
bowenlan-amzn, line 42: not sure about what this test is checking, would you explain a bit

src/main/kotlin/org/opensearch/commons/replication/action/ReplicationActions.kt
```diff
@@ -0,0 +1,31 @@
...
+object ReplicationActions {
+
+    /**
+     * Action names for stopping replication
+     * STOP_REPLICATION_ACTION_NAME: action used for _stop REST API
+     * UNFOLLOW_REPLICATION_ACTION_NAME: internal action used for inter-plugin communication i.e. by ism to invoke stop
+     * replication.
+     */
+    const val STOP_REPLICATION_ACTION_NAME = "indices:admin/plugins/replication/index/stop"
+    const val UNFOLLOW_REPLICATION_ACTION_NAME = "indices:admin/plugins/replication/index/unfollow"
```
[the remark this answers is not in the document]
aggarwalShivani: Yeah I agreee 😅 
But I was wondering if stop_from_ism would limit it to being only from ism? and what if some other plugin in future intends to invoke stop-replication using these libs.  🤔  
Would something like stop_from_transport / transport_stop / stop_from_plugin work?

src/test/kotlin/org/opensearch/commons/replication/ReplicationPluginInterfaceTests.kt
aggarwalShivani: Sure. I've written this test on similar lines as done in [NotificationsPluginInterfaceTests.kt]( https://github.com/opensearch-project/common-utils/blob/main/src/test/kotlin/org/opensearch/commons/notifications/NotificationsPluginInterfaceTests.kt#L63) and [AlertingPluginInterfaceTests.kt](https://github.com/opensearch-project/common-utils/blob/20f9ef6a315c0c51279cdcbe02e27597e3399afa/src/test/kotlin/org/opensearch/commons/alerting/AlertingPluginInterfaceTests.kt#L58)

- An AcknowledgedResponse response object is created with a value of true (expected response). 
- Have mocked the client object such that, when stopReplication() invokes client.execute(), it would trigger the onResponse method of the ActionListener with the successful **response** object created earlier.
- The test runs stopReplication() and verifies that the onResponse() method of the listener was called exactly once with the provided response object as an argument.
- This is to simulate a successful execution and verifies that the stopReplication method correctly handles the response.

src/main/kotlin/org/opensearch/commons/replication/action/ReplicationActions.kt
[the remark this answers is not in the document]
bowenlan-amzn: yeah, better to be intuitive for our user.

Looking at all the existing ones, I am thinking `indices:transport/plugins/replication/index/stop`
Hopefully transport is a common word for OpenSearch user.

https://github.com/opensearch-project/cross-cluster-replication/blob/d6e7636d42ef87cd483f7439d8a2cf43bf397727/HANDBOOK.md?plain=1#L108-L121

```
# Index Level Permissions

indices:admin/close
indices:admin/close[s]
indices:admin/create
indices:admin/mapping/put
indices:admin/open
indices:admin/plugins/replication/index/start
indices:admin/plugins/replication/index/stop
indices:data/read/plugins/replication/file_metadata
indices:data/write/index
indices:data/write/plugins/replication/changes
indices:data/write/replication
indices:monitor/stats
```

@cwperks to give some suggestion about the naming of the new transport action
This new transport action being added here is only supposed to be called from ISM plugin, and doesn't have any REST API associated with it. ISM user will need this permission to do the stop replication action using ISM.

src/test/kotlin/org/opensearch/commons/replication/ReplicationPluginInterfaceTests.kt
bowenlan-amzn: Thanks for the explain

What I feel weird is we mock the client.execute to do listener.onResponse
Then we call stopReplication which just does client.execute, and verify listener.onResponse called once

In the end, it seems just testing stopReplication does client.execute exactly once 🤔

src/main/kotlin/org/opensearch/commons/replication/action/ReplicationActions.kt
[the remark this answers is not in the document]
cwperks: @bowenlan-amzn `indices:transport` is not one of the valid action prefixes tracked in [core](https://github.com/opensearch-project/OpenSearch/blob/6b8b3efe01a62c221f308a2e3b019d75a7f5ad8a/server/src/main/java/org/opensearch/transport/TransportService.java#L1097-L1112). You can introduce a new action prefix by opening a PR in core to add to that list.

I see there is `indices:internal` in the list of prefixes. Is that prefix reserved for actions defined in OpenSearch core, or is it possible for a plugin to use that prefix as well?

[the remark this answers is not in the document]
bowenlan-amzn: Thanks!

`indices:internal` makes sense here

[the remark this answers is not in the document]
aggarwalShivani: I will try with indices:internal/plugins/replication/index/stop then 🙂 
Thank you @cwperks  and @bowenlan-amzn for these inputs.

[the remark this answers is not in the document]
aggarwalShivani: Hi, 
Thanks, I've gone with the same as suggested.  
> INTERNAL_STOP_REPLICATION_ACTION_NAME = "indices:internal/plugins/replication/index/stop"

src/main/kotlin/org/opensearch/commons/replication/ReplicationPluginInterface.kt
aggarwalShivani: - BaseResponse is an abstract class and extends ActionResponse. For alerting/notification actions, separate response classes are created (ex. GetMonitorResponse, SendNotificationResponse) that extend BaseResponse and add their own functionality on top. 
- AcknowledgedResponse is a different subclass of ActionResponse and in the ccr plugin, the transport action ([TransportStopIndexReplicationAction](https://github.com/opensearch-project/cross-cluster-replication/blob/main/src/main/kotlin/org/opensearch/replication/action/stop/TransportStopIndexReplicationAction.kt)) expects a listener of type AcknowledgedResponse. 
- Also, in ism, on invoking this action, we want to access properties like response.isAcknowledged (which is not the case in [notifications ](https://github.com/opensearch-project/index-management/blob/main/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/notification/AttemptNotificationStep.kt)action).

For this reason, I've used AcknowledgedResponse here.

aggarwalShivani:
Hi, 
For this change, some CI failures are due to AcknowledgedResponse moving to org.opensearch.action.support.clustermanager. Will repush once the integration issue with common-utils, ccr, ism on main branch is resolved
As discussed with @bowenlan-amzn, raised backport PR on 2.x branch (#789).

aggarwalShivani:
> Hi, For this change, some CI failures are due to AcknowledgedResponse moving to org.opensearch.action.support.clustermanager. Will repush once the integration issue with common-utils, ccr, ism on main branch is resolved As discussed with @bowenlan-amzn, raised backport PR on 2.x branch (#789).

Hi @bowenlan-amzn 

The integration issue of common-utils and ccr is resolved with the merge of this [change](https://github.com/opensearch-project/cross-cluster-replication/pull/1504) in ccr. 
ISM continues to remain blocked though until this [issue](https://github.com/opensearch-project/index-management/issues/1373) is resolved. 

Meanwhile, as the changes for the [stop-replication feature](https://github.com/opensearch-project/index-management/issues/726) for both common-utlis and ccr are ready, requesting your review on this.
Backport PR is also ready for review #789

bowenlan-amzn (approved):

src/main/kotlin/org/opensearch/commons/replication/ReplicationPluginInterface.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.opensearch.commons.replication.action.StopIndexReplicationRequest
+import org.opensearch.commons.utils.recreateObject
+import org.opensearch.core.action.ActionListener
+import org.opensearch.core.action.ActionResponse
+import org.opensearch.core.common.io.stream.Writeable
+
+/**
+ * Transport action plugin interfaces for the cross-cluster-replication plugin.
+ */
+open class ReplicationPluginInterface {
```
bowenlan-amzn, line 20:
```suggestion
object ReplicationPluginInterface {

```

bowenlan-amzn: Feels like the methods here should just be static methods.

aggarwalShivani: Thanks. I've made this change now 👍

bowenlan-amzn (approved):

PR #758, opened 2024-12-17
Monitor model changed to add an optional fanoutEnabled field

riysaxen-amzn:
### Description
Detectors configured with aggregation sigma rules should ideally generate only one alert in one execution when a set of documents match the rule’s condition. But currently there are duplicate alerts being generated when aggregate sigma rules are matched.
This PR adds a field in monitor  `fanoutEnabled` (optional field), default value of this flag will be true. When detector is created/updated, the value of this field will be set to false in case chained findings doc level monitor, this will ensure that fanout approach will be disabled when executing doc level monitor

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
@@ -44,7 +44,8 @@ data class Monitor(
     val dataSources: DataSources = DataSources(),
     val deleteQueryIndexInEveryRun: Boolean? = false,
     val shouldCreateSingleAlertForFindings: Boolean? = false,
-    val owner: String? = "alerting"
+    val owner: String? = "alerting",
+    val fanoutEnabled: Boolean? = true,
 ) : ScheduledJob {
 
     override val type = MONITOR_TYPE
@@ -114,6 +115,7 @@ data class Monitor(
         },
         deleteQueryIndexInEveryRun = sin.readOptionalBoolean(),
         shouldCreateSingleAlertForFindings = sin.readOptionalBoolean(),
+        fanoutEnabled = sin.readOptionalBoolean(),
         owner = sin.readOptionalString()
     )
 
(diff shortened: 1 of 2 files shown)


eirsep (approved):

src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
```diff
@@ -44,7 +44,8 @@ data class Monitor(
     val dataSources: DataSources = DataSources(),
     val deleteQueryIndexInEveryRun: Boolean? = false,
     val shouldCreateSingleAlertForFindings: Boolean? = false,
-    val owner: String? = "alerting"
+    val owner: String? = "alerting",
+    val fanoutEnabled: Boolean? = true
```
AWSHurneyt, line 48: This attribute would be specific to doc level monitors, right? Should it instead be in the doc level monitor input then?
https://github.com/opensearch-project/common-utils/blob/main/src/main/kotlin/org/opensearch/commons/alerting/model/DocLevelMonitorInput.kt

src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
```diff
@@ -175,6 +177,7 @@ data class Monitor(
         builder.field(DATA_SOURCES_FIELD, dataSources)
         builder.field(DELETE_QUERY_INDEX_IN_EVERY_RUN_FIELD, deleteQueryIndexInEveryRun)
         builder.field(SHOULD_CREATE_SINGLE_ALERT_FOR_FINDINGS_FIELD, shouldCreateSingleAlertForFindings)
+        builder.field(FAN_OUT_ENABLED_FIELD, fanoutEnabled)
```
AWSHurneyt, line 180: If we decide to keep this attribute at the monitor level instead of moving it to the doc level monitor input class, we should probably only include the field in the `toXContent` output for doc level monitors to avoid confusion.

src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
riysaxen-amzn: addressed

AWSHurneyt (approved):

eirsep (approved):
Unit test not added

riysaxen-amzn:
UTs will add as a follow up

PR #770, opened 2025-01-16
API to enable/disable a monitor

vikhy-aws:
### Description
This PR adds request and response classes to common-utils to create an API that enables or disables the monitor using the following:

`PUT _plugins/_alerting/monitors/{monitor_id}/enable`
`PUT _plugins/_alerting/monitors/{monitor_id}/disable`

### Related Issues
Resolves https://github.com/opensearch-project/alerting/issues/1058

### Check List


src/main/kotlin/org/opensearch/commons/alerting/action/ToggleMonitorRequest.kt
```diff
@@ -0,0 +1,52 @@
...
+    override fun writeTo(out: StreamOutput) {
+        out.writeString(monitorId)
+        out.writeBoolean(enabled)
+        out.writeLong(seqNo)
+        out.writeLong(primaryTerm)
+        out.writeEnum(method)
+    }
+
+    override fun validate(): ActionRequestValidationException? {
+        return null
```
riysaxen-amzn, line 50: what's this function about?

vikhy-aws: It's an abstract method defined in the ActionRequest class, so need to provide an implementation here. Based on it's usage in other classes, it is used to validate the data members, which is not required in this case.

src/main/kotlin/org/opensearch/commons/alerting/action/ToggleMonitorResponse.kt
```diff
@@ -0,0 +1,64 @@
...
+        version: Long,
+        seqNo: Long,
+        primaryTerm: Long,
+        monitor: Monitor
+    ) : super() {
+        this.id = id
+        this.version = version
+        this.seqNo = seqNo
+        this.primaryTerm = primaryTerm
+        this.monitor = monitor
```
AWSHurneyt, line 33: We may want to consider excluding the `monitor` from the response. Security analytics creates per document monitors that can have thousands of queries (a.k.a., rules) in them. That can make the response size pretty big, which may be extraneous for this kind of API. Perhaps we can instead just return the updated `enabled` value, instead of the entire monitor? We can discuss with the team.

AWSHurneyt (commented):
Are any changes needed in the alerting plugin interface? It may be necessary for the security analytics plugin to enable/disable monitors.
https://github.com/opensearch-project/common-utils/blob/main/src/main/kotlin/org/opensearch/commons/alerting/AlertingPluginInterface.kt

PR #801, opened 2025-03-17
Escape/Unescape pipe UserInfo in ThreadContext

shikharj05:
### Description
Escape pipe while setting UserInfo in ThreadContext, Unescape pipe while reading UserInfo from ThreadContext

### Related Issues
https://github.com/opensearch-project/security/issues/2756

### Check List
- [] New functionality includes testing.
- [NA] New functionality has been documented.
- [NA] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [X] Commits are signed per the DCO using `--signoff`.
- [NA] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/src/main/java/org/opensearch/commons/InjectSecurity.java
+++ b/src/main/java/org/opensearch/commons/InjectSecurity.java
@@ -8,6 +8,7 @@ package org.opensearch.commons;
 import static org.opensearch.commons.ConfigConstants.INJECTED_USER;
 import static org.opensearch.commons.ConfigConstants.OPENSEARCH_SECURITY_INJECTED_ROLES;
 import static org.opensearch.commons.ConfigConstants.OPENSEARCH_SECURITY_USE_INJECTED_USER_FOR_PLUGINS;
+import static org.opensearch.commons.authuser.Utils.escapePipe;
 
 import java.util.List;
 import java.util.StringJoiner;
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/commons/InjectSecurity.java
```diff
@@ -143,13 +144,16 @@ public void injectUserInfo(final User user) {
                 );
             return;
         }
+
         StringJoiner joiner = new StringJoiner("|");
```
cwperks, line 148: Curious to know where this serialization logic is used because I thought security was the only place where the user object would be serialized into the ThreadContext. Is this used for interplay between plugins in certain contexts?

cwperks: While it may make sense to centralize this logic to prevent duplication, long term I'd like to remove a lot of security logic from common-utils including this.

lezzago (dismissed):

src/main/java/org/opensearch/commons/InjectSecurity.java
lezzago: > Curious to know where this serialization logic is used because I thought security was the only place where the user object would be serialized into the ThreadContext. Is this used for interplay between plugins in certain contexts?

Yes, this is used for plugins to communicate with one another.



> While it may make sense to centralize this logic to prevent duplication, long term I'd like to remove a lot of security logic from common-utils including this.

I agree, this is really needed, but that might require some security features in core.

cwperks: I have an RFC open in the security repo that outlines 4 security features that need to be built in order for plugins to migrate off of using common-utils: https://github.com/opensearch-project/security/issues/5052

First, the security repo needs to support existing use cases in a model that's opt-in for plugins so that they can adopt the new mechanisms and create migration paths. Particularly, a lot of plugins store a copy of a user in their own system indices and eventually I think we should remove the copies but mechanisms need to be built to support it first.

cwperks (approved):

src/main/java/org/opensearch/commons/authuser/User.java
```diff
@@ -166,24 +167,26 @@ public static User parse(final String userString) {
             return null;
         }
 
-        String[] strs = userString.split("\\|");
```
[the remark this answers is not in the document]
shikharj05: This is to ensure correct parsing with multiple pipes as well. e.g. given that user-name can contain multiple pipes.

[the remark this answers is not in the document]
shikharj05: I have added a test case to verify the same ``testParseUserStringWithMultiplePipesInUserName``

cwperks (approved):

lezzago (approved):

PR #827, opened 2025-05-12
Update user parsing to include custom attribute names

markdboyd:
### Description

As part of https://github.com/opensearch-project/alerting/issues/1829, I found that the user parsing in this library, which is used by the alerting plugin, does not parse custom attribute names from a user string. 

The current implementation makes sense because the security plugin does not currently include custom attribute names in the user string in the thread context: https://github.com/opensearch-project/security/blob/dfe16b004a6bb1c8d2bc50ac0c0f76780a3519cb/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L276-L290. 

However, I plan to update the security plugin to include custom attribute names in the thread context. Ultimately, the goal of this change and other related changes is to make sure that user custom attributes are available when the alerting plugin executes monitor queries, so that DLS parameter substitution will work correctly.

### Related Issues

Related to https://github.com/opensearch-project/alerting/issues/1829

### Check List
- [ ] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/src/main/java/org/opensearch/commons/authuser/User.java
+++ b/src/main/java/org/opensearch/commons/authuser/User.java
@@ -178,6 +178,7 @@ final public class User implements Writeable, ToXContent {
         List<String> backendRoles = new ArrayList<>();
         List<String> roles = new ArrayList<>();
         String requestedTenant = null;
+        List<String> customAttNames = new ArrayList<>();
 
         if ((strs.length > 1) && !Strings.isNullOrEmpty(strs[1])) {
             backendRoles.addAll(Arrays.stream(strs[1].split(",")).map(Utils::unescapePipe).toList());
@@ -188,7 +189,10 @@ final public class User implements Writeable, ToXContent {
         if ((strs.length > 3) && !Strings.isNullOrEmpty(strs[3])) {
             requestedTenant = unescapePipe(strs[3].trim());
         }
-        return new User(userName, backendRoles, roles, Arrays.asList(), requestedTenant);
+        if ((strs.length > 4) && !Strings.isNullOrEmpty(strs[4])) {
+            customAttNames.addAll(Arrays.stream(strs[4].split(",")).map(Utils::unescapePipe).toList());
+        }
+        return new User(userName, backendRoles, roles, customAttNames, requestedTenant);
     }
 
     @Override
--- a/build.gradle
+++ b/build.gradle
@@ -85,12 +85,12 @@ dependencies {
     testImplementation "org.opensearch.test:framework:${opensearch_version}"
     testImplementation "org.jetbrains.kotlin:kotlin-test:${kotlin_version}"
     testImplementation "org.mockito:mockito-core:3.10.0"
-    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.7.2'
+    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.11.4'
     testImplementation 'org.mockito:mockito-junit-jupiter:3.10.0'
     testImplementation "com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0"
     testImplementation "com.cronutils:cron-utils:9.1.6"
     testImplementation "commons-validator:commons-validator:1.7"
-    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.2'
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.11.4'
 
     ktlint "com.pinterest:ktlint:0.47.1"
 }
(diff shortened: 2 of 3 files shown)


src/main/java/org/opensearch/commons/authuser/User.java
```diff
@@ -188,7 +189,10 @@ public static User parse(final String userString) {
         if ((strs.length > 3) && !Strings.isNullOrEmpty(strs[3])) {
             requestedTenant = unescapePipe(strs[3].trim());
         }
```
cwperks, line 191: We should keep the `strs.length > n` checks for backwards compatibility, but once all nodes are on a version of opensearch where the user is serialized to 5 values then we'd only need one top-level check.

This block will also need a special condition `if ("null".equals(strs[3])) { requestedTenant = null; }`

src/main/java/org/opensearch/commons/authuser/User.java
```diff
@@ -188,7 +189,10 @@ public static User parse(final String userString) {
...
-        return new User(userName, backendRoles, roles, Arrays.asList(), requestedTenant);
+        if ((strs.length > 4) && !Strings.isNullOrEmpty(strs[4])) {
+            customAttNames.addAll(Arrays.stream(strs[4].split(",")).map(Utils::unescapePipe).toList());
```
cwperks, line 193: This block needs to be updated to do base64 deserialization since it would be base64 encoded from the security plugin. 

Let's also change `customAttNames` to a Map bc we should stores keys and values, not only the names of the attributes.

Any caller that needs to attribute names can get the keySet of the map

cwperks: Change it to a map will require additional changes for serializing the User to streamOutput, reading from streamInput and also serialization/deserializing to and from XContent.

cwperks:
I see that this repo has customAttrNames, however, I'm not sure how its used because the security repo does not serialize it. I think its safe to change the data structure to a Map and include the values as well. I created this change on my fork https://github.com/cwperks/common-utils/pull/1, but haven't adequately done any e2e test with your changes on the security repo yet.

cwperks (commented):
Thank you @markboyd.

I'd like to look into 2 more things:

- Let's see about adding a test for the changes in `public static User parse(XContentParser parser)`
- Potentially copying over any tests from the security repo for the Base64Helper class

src/main/java/org/opensearch/commons/authuser/User.java
```diff
@@ -141,10 +143,13 @@ public static User parse(XContentParser parser) throws IOException {
                         roles.add(parser.text());
                     }
                     break;
-                case CUSTOM_ATTRIBUTE_NAMES_FIELD:
-                    ensureExpectedToken(XContentParser.Token.START_ARRAY, parser.currentToken(), parser);
-                    while (parser.nextToken() != XContentParser.Token.END_ARRAY) {
-                        customAttNames.add(parser.text());
+                case CUSTOM_ATTRIBUTES_FIELD:
+                    ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
```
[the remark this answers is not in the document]
markdboyd: OK, this is complete here: https://github.com/opensearch-project/common-utils/pull/827/commits/3aa0ab0df00a46f82e3e49b2ab9e50fda67d12f6

src/test/java/org/opensearch/commons/authuser/util/Base64HelperTest.java
markdboyd, line 1: This file is copied from https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/support/Base64HelperTest.java

I excluded these tests 

- https://github.com/opensearch-project/security/blob/f548f51b6cecfad3717da6efd118151c6212e91e/src/test/java/org/opensearch/security/support/Base64HelperTest.java#L67-L71 - I don't think the `User` class in common-utils is Serializable, while the `User` class in the `security` plugin is: https://github.com/opensearch-project/security/blob/f548f51b6cecfad3717da6efd118151c6212e91e/src/main/java/org/opensearch/security/user/User.java#L59C30-L59C42

https://github.com/opensearch-project/security/blob/f548f51b6cecfad3717da6efd118151c6212e91e/src/test/java/org/opensearch/security/support/Base64HelperTest.java#L73-L77 - The SourceFieldsContext class under test doesn't exist in common-utils

cwperks (dismissed):
Changes LGTM! Thank you @markdboyd 

For the maintainers of this repo, we need to merge this and https://github.com/opensearch-project/security/pull/5491 at the same time

cwperks (approved):

DarshitChanpura (approved):

cwperks (dismissed):

cwperks (approved):

cwperks (approved):

cwperks (approved):

PR #940, opened 2026-04-16
PPL Alerting: Adding PPL Related Models

toepkerd:
### Description
Changes to support PPL Alerting behind existing (V1) Alerting APIs, and to support PPL Alerting using stateful Alerts instead of stateless Alerts.

PPL Monitors will be run by the QueryLevelMonitorRunner (much like AD and Cluster Metrics Monitors), so these changes prepare for that as well.

### Check List


src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
```diff
@@ -102,7 +167,9 @@ data class Alert(
         schemaVersion: Int = NO_SCHEMA_VERSION,
         executionId: String? = null,
         workflowId: String? = null,
-        clusters: List<String>? = null
+        clusters: List<String>? = null,
+        pplQuery: String? = null,
+        pplQueryResults: List<Map<String, Any?>> = listOf()
```
lezzago: Do we want to generalize these fields so it could be the any query for also normal query level monitors and what api for cluster metrics? Also the same for the results.

src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
```diff
@@ -264,6 +292,10 @@ data class Monitor(
         const val OWNER_FIELD = "owner"
         val MONITOR_TYPE_PATTERN = Pattern.compile("[a-zA-Z0-9_]{5,25}")
 
+        // hard, nonadjustable limits for PPL Alerting
+        const val ALERTING_MAX_NAME_LENGTH = 30 // max length of any name for monitors, triggers, notif actions, etc
```
eirsep, line 331: why are we adding this as part of PPL monitor type addition?? 

how will this be backward compatible?

src/main/kotlin/org/opensearch/commons/alerting/model/PPLTrigger.kt
```diff
@@ -0,0 +1,335 @@
...
+ * @property conditionType The type of condition to evaluate.
+ *               Can be either [ConditionType.NUMBER_OF_RESULTS] or [ConditionType.CUSTOM].
+ * @property numResultsCondition The comparison operator for NUMBER_OF_RESULTS conditions. Required if using NUMBER_OF_RESULTS conditions,
+ *                               required to be null otherwise.
+ * @property numResultsValue The threshold value for NUMBER_OF_RESULTS conditions. Required if using NUMBER_OF_RESULTS conditions,
+ *                           required to be null otherwise.
+ * @property customCondition A custom condition expression. Required if using CUSTOM conditions,
+ *                           required to be null otherwise.
+ *
+ * @opensearch.experimental
```
eirsep: experimental is present intermittently? either remove everywhere or add where missing

src/main/kotlin/org/opensearch/commons/alerting/model/QueryLevelTriggerRunResult.kt
```diff
@@ -47,13 +49,15 @@ open class QueryLevelTriggerRunResult(
         return builder
             .field("triggered", triggered)
             .field("action_results", actionResults as Map<String, ActionRunResult>)
+            .field("ppl_query_results", pplCustomQueryResults)
```
eirsep, line 57: why are the serialized field name not "ppl_sql" prefixed similar to class and variable names.
let's remove sql substring everywhere if possible or make the serialized names consistent.

@lezzago  thoughts?

src/main/kotlin/org/opensearch/commons/alerting/model/MonitorRunResult.kt
```diff
@@ -108,6 +124,11 @@ data class InputRunResults(
         for (map in results) {
             out.writeMap(map)
         }
+        out.writeVInt(pplBaseQueryResults.size)
```
eirsep: the readFrom and writeTo() methods don't seem to be consistent... did you add this serde roundtrip in test coverage with new fields and without new fields? if yes, plz respond to this comment with the test methods

src/test/kotlin/org/opensearch/commons/alerting/TestHelpers.kt
```diff
@@ -681,7 +746,7 @@ fun createCorrelationAlertTemplateArgs(correlationAlert: CorrelationAlert): Map<
 }
 
 fun randomInputRunResults(): InputRunResults {
-    return InputRunResults(listOf(), null)
+    return InputRunResults(listOf(), null, null, listOf(), 5L)
```
eirsep: plz dont add nullable stuff alone. kindly add proper test coverage for inputrunresults serde

eirsep (changes requested):
There is a general mistake in how we are doing readFrom() and writeTo() - basically the writeable serialization.
old classes being deserialized or serialized will fail over the wire since they expect these new PPL query results or other PPL specific changes. you need to add some version checks before expecting these to be avaiablle and serialized or deserialized.

src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
```diff
@@ -87,12 +148,16 @@ data class Alert(
...
+        pplQuery = null,
+        pplQueryResults = listOf()
     )
 
+    // constructor for Alerts from QueryLevelMonitorRunner
+    // this monitor runner runs Query-level, Cluster Metrics, and PPL Monitors
     constructor(
         monitor: Monitor,
-        trigger: QueryLevelTrigger,
+        trigger: Trigger,
```
eirsep, line 165: why are we changing this?  how or why is this in scope of PPL models related changes?

src/test/kotlin/org/opensearch/commons/alerting/TestHelpers.kt
```diff
@@ -325,6 +357,30 @@ fun randomChainedAlertTrigger(
     )
 }
 
+fun randomPPLSQLTrigger(
+    id: String = UUIDs.base64UUID(),
+    name: String = RandomStrings.randomAsciiLettersOfLength(Random(), 10),
+    severity: String = "1",
+    actions: List<Action> = mutableListOf(),
+    conditionType: PPLSQLTrigger.ConditionType = PPLSQLTrigger.ConditionType.NUMBER_OF_RESULTS,
```
eirsep: - No test for CUSTOM condition type trigger (all random triggers default to NUMBER_OF_RESULTS)

src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
toepkerd: The List<Map<String, Any?>> typing would work for both DSL and PPL query results but we'd need to be mindful that the format of the contents are not at all the same. For Query-Level Monitor runs, the List<Map<String, Any?>> would contain exactly 1 element of the DSL query results in its usual format. For PPL Monitor runs, the List<Map<String, Any?>> would be of this format:
```
 List<Map<String, Any?>> = listOf(                                                                                                                                                   
      mapOf(                                                                                                                                                                          
          "region" to "us-east-1",                                                                                                                                                    
          "max_price" to 450.0,                                                                                                                                                       
          "avg_latency" to 120.5                                                                                                                                                      
      ),                                                                                                                                                                              
      mapOf(                                                                                                                                                                          
          "region" to "us-west-2",                                                                                                                                                    
          "max_price" to 380.0,                                                                                                                                                       
          "avg_latency" to 95.3                                                                                                                                                       
      ),                                                                                                                                                                              
      mapOf(                                                                                                                                                                          
          "region" to "eu-west-1",                                                                                                                                                    
          "max_price" to 520.0,                                                                                                                                                       
          "avg_latency" to 145.8                                                                                                                                                      
      )                                                                                                                                                                               
  ) 
```

Seeing that the query results in an Alert are only ever meant to be readable by a user that calls Get Alerts, and not accessed with an `alert.field` call, the generalization should be fine.

src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
toepkerd: This length check is from a previous Security review of PPL Alerting behind V2 APIs. The name length check is only done on PPL Monitors, it's not going to be done on existing Monitor types.

src/main/kotlin/org/opensearch/commons/alerting/model/QueryLevelTriggerRunResult.kt
toepkerd: Fair callout that there is inconsistency between calling it a "PPL" thing or a "PPLSQL" thing. The only trouble with calling everything "PPLSQL" is that the name of the PPL Trigger type has to be `ppl_trigger` and not `ppl_sql_trigger` since that would mean users creating PPL Monitors through Dev Tools or programmatically would need to specify that name `ppl_sql_trigger`, even though we are seeking to hide potential SQL support for now. Our options are:
1. Remove SQL substring from everything, and then live with the fact that if PPL Monitors support SQL queries in the future, they will continue to be called just PPL Monitors and PPL Triggers
2. Add SQL substring to everything, and then let the construct names hint at future SQL support

I'm leaning toward option 1

src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
toepkerd: Because PPL Monitors will not have their own dedicated `PPLMonitorRunner`. They will run on `QueryLevelMonitorRunner`, much like AD and Cluster Metrics Monitors do. The thing is PPL Triggers need to be their own type because they are fundamentally different from QueryLevelTriggers. As such, the type of `trigger` had to be generalized to `Trigger` interface so that it can handle either `QueryLevelTrigger` or `PPLTrigger`.

This constructor is called by `composeQueryLevelAlert` in Alerting. There are validations in Alerting that the trigger passed into this constructor is only ever type `QueryLevelTrigger` or `PPLTrigger`.

src/main/kotlin/org/opensearch/commons/alerting/model/MonitorRunResult.kt
toepkerd: Only a roundtrip test with new fields, and even then one of the new fields is just null as you pointed out in another comment: `WriteableTests.test inputrunresult as stream`. Will improve the test coverage to not use a null PPL query results field, and to test with + without new PPL fields separately.

src/test/kotlin/org/opensearch/commons/alerting/TestHelpers.kt
toepkerd: Will change to use an actual list of maps

src/main/kotlin/org/opensearch/commons/alerting/model/QueryLevelTriggerRunResult.kt
```diff
@@ -18,7 +18,8 @@ open class QueryLevelTriggerRunResult(
     override var triggerName: String,
     open var triggered: Boolean,
     override var error: Exception?,
-    open var actionResults: MutableMap<String, ActionRunResult> = mutableMapOf()
+    open var actionResults: MutableMap<String, ActionRunResult> = mutableMapOf(),
+    open var pplCustomQueryResults: List<Map<String, Any?>> = listOf()
```
lezzago, line 23: change this to be generic

src/main/kotlin/org/opensearch/commons/alerting/model/QueryLevelTriggerRunResult.kt
lezzago: Lets remove SQL since if we add SQL support, that would be a new language and we dont need to show that SQL and PPL are tied together in one plugin.

src/main/kotlin/org/opensearch/commons/alerting/model/QueryLevelTriggerRunResult.kt
toepkerd: I actually believe keeping it PPL is the better approach for this field specifically. Query-Level Monitors run their query once, then apply painless trigger execution on those results. PPL Monitors with a custom trigger conditions append the condition to the base PPL query. If a PPL Monitor has multiple custom triggers, each Trigger will be running a different query. 

In other words, Query-Level Monitors run the query per Monitor execution. PPL Monitors run the query per Trigger execution. Having query results specifically in the Trigger Run Result is unique to PPL Monitors.

lezzago: > Query-Level Monitors run their query once, then apply painless trigger execution on those results. PPL Monitors with a custom trigger conditions append the condition to the base PPL query. If a PPL Monitor has multiple custom triggers, each Trigger will be running a different query.

Should we change this so its similar to query level monitors? First run the base query and then run the trigger condition query ontop of the PPL query? This is to help search costs. This may be more complex or if we can have this is a temp tables for the results and then run the different trigger queries against it, that would be best. Though this would depend on the feasibility with PPL features.

src/main/kotlin/org/opensearch/commons/alerting/model/PPLTrigger.kt
toepkerd: Removing

src/main/kotlin/org/opensearch/commons/alerting/model/QueryLevelTriggerRunResult.kt
toepkerd: This option has been explored and deemed not possible. PPL must support the `makeresults` key word to make PPL Monitors run in the 2-step process you describe above (1. run base query 2. apply trigger condition to results). Even if this key word was supported, custom condition triggers each run their own query to evaluate trigger condition (base query + custom condition).

That said, optimizations were already done post load testing to run just the base query once, so that number of results triggers need not rerun the base query redundantly. This optimization cannot be done for custom triggers though.

src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
```diff
@@ -87,6 +90,27 @@ data class Monitor(
...
+            // this is a new check for PPL Alerting specifically, PPL Monitors will enforce
+            // a max name length, but it won't be enforced on other Monitor types to avoid
+            // adding a breaking change
+            require(this.name.length <= ALERTING_MAX_NAME_LENGTH) {
+                "Monitor name too long, length must be at most $ALERTING_MAX_NAME_LENGTH."
+            }
+
+            // this is a new check for PPL Alerting specifically, other Monitor types allow
+            // themselves to be created without any Triggers
+            require(this.triggers.isNotEmpty()) { "PPL Monitor must include at least 1 trigger." }
```
eirsep, line 112: shouldn't there be another check that PPL can have exactly 1 trigger?

src/main/kotlin/org/opensearch/commons/alerting/model/MonitorRunResult.kt
```diff
@@ -92,12 +93,28 @@ data class MonitorRunResult<TriggerResult : TriggerRunResult>(
 data class InputRunResults(
     val results: List<Map<String, Any>> = listOf(),
     val error: Exception? = null,
-    val aggTriggersAfterKey: MutableMap<String, TriggerAfterKey>? = null
+    val aggTriggersAfterKey: MutableMap<String, TriggerAfterKey>? = null,
+    val pplBaseQueryResults: List<Map<String, Any?>> = listOf(),
```
eirsep, line 100: why "base" ?

toepkerd: The base PPL query is what the PPL Monitor is configured with. Custom condition triggers append to this base query to run their own queries.

src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
toepkerd: PPL Monitors can have multiple Triggers. They are, however, only allowed to have 1 Input.

src/main/kotlin/org/opensearch/commons/alerting/model/PPLInput.kt
```diff
@@ -0,0 +1,110 @@
...
+        val XCONTENT_REGISTRY = NamedXContentRegistry.Entry(
+            Input::class.java,
+            ParseField(PPL_INPUT_FIELD),
+            CheckedFunction { parseInner(it) }
+        )
+
+        @JvmStatic
+        @Throws(IOException::class)
+        fun parseInner(xcp: XContentParser): PPLInput {
+            lateinit var query: String
```
eirsep: why lateinit??

src/test/kotlin/org/opensearch/commons/alerting/TestHelpers.kt
```diff
@@ -681,7 +761,26 @@ fun createCorrelationAlertTemplateArgs(correlationAlert: CorrelationAlert): Map<
 }
 
 fun randomInputRunResults(): InputRunResults {
-    return InputRunResults(listOf(), null)
+    return InputRunResults(listOf(), null, null)
+}
+
+fun randomInputRunResultsWithPPLFields(): InputRunResults {
+    return InputRunResults(
+        listOf(),
```
eirsep: what is this field? plz don't pass empty. this won't test serde roundtrip extensively

src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
```diff
@@ -337,6 +417,16 @@ data class Alert(
...
+        query = if (sin.version.onOrAfter(Version.V_3_7_0)) {
+            sin.readOptionalString()
+        } else {
+            null
+        },
+        queryResults = if (sin.version.onOrAfter(Version.V_3_7_0)) {
+            sin.readList { input -> suppressWarning(input.readMap()) }
+        } else {
+            listOf()
+        },
```
lezzago, line 429: The check should not be required since optional handles BWC.

src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
```diff
@@ -380,6 +470,12 @@ data class Alert(
         out.writeOptionalString(executionId)
         out.writeStringCollection(associatedAlertIds)
         out.writeOptionalStringArray(clusters?.toTypedArray())
+        if (out.version.onOrAfter(Version.V_3_7_0)) {
```
lezzago, line 473: the check shouldnt be required

src/main/kotlin/org/opensearch/commons/alerting/model/MonitorRunResult.kt
lezzago: should we add comments to make this clear?

src/main/kotlin/org/opensearch/commons/alerting/model/PPLTrigger.kt
```diff
@@ -0,0 +1,333 @@
...
+        const val NUM_RESULTS_CONDITION_FIELD = "num_results_condition"
+        const val NUM_RESULTS_VALUE_FIELD = "num_results_value"
+        const val CUSTOM_CONDITION_FIELD = "custom_condition"
+
+        // hard, nonadjustable limits
+        const val NOTIFICATIONS_ID_MAX_LENGTH = 512 // length limit for notifications channel custom ID at channel creation time
+
+        // regular expression for validating that a string contains
+        // only valid chars (letters, numbers, -, _)
+        private val validCharsRegex = """^[a-zA-Z0-9_-]+$""".toRegex()
```
lezzago: is this really needed? Its for the destination id which would be the notification channel id. We dont create or generate an id from this plugin. Does it make sense for this validation?

src/main/kotlin/org/opensearch/commons/alerting/model/QueryLevelTriggerRunResult.kt
lezzago: Lets leave this as an open github issue with PPL plugin for the feature request for performance optimizations. Make sure that `pplCustomQueryResults` is not customer facing so we could change this variable name down the road.

src/main/kotlin/org/opensearch/commons/alerting/model/PPLInput.kt
toepkerd: No specific reason for lateinit, can change to initialize with null

src/test/kotlin/org/opensearch/commons/alerting/TestHelpers.kt
toepkerd: It's the `results` field for the existing Monitor types. The idea is that DSL Monitors would only ever use the `results` field, and PPL Monitors will only ever use the `pplBaseQueryResults` field. For testing purposes, though, there's no harm in actually populating this list.

src/main/kotlin/org/opensearch/commons/alerting/model/MonitorRunResult.kt
toepkerd: Will add clarifying comment

src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
toepkerd: writeOptional adds a bool to the bytestream and readOptional expects it. The optional()s don't check for presence/absence of bytes

src/main/kotlin/org/opensearch/commons/alerting/model/PPLTrigger.kt
toepkerd: This change was requested during a security review of PPL Alerting. The reasoning is that Channel IDs are user input (e.g. in CreateMonitor API) that must be sanitized.

src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
toepkerd: writeOptional adds a bool to the bytestream and readOptional expects it. The optional()s don't check for presence/absence of bytes

src/test/kotlin/org/opensearch/commons/alerting/TestHelpers.kt
toepkerd: Upcoming change will only populate this `results` list, but not touch existing randomInputRunResults() helpers that return a null `aggTriggersAfterKey` and `exception`. `aggTriggersAfterKey` is intentionally left out of serde ([reference](https://github.com/toepkerd/common-utils/blob/2f9a97cb0ff0482925adada60177c3f41c7df384/src/main/kotlin/org/opensearch/commons/alerting/model/MonitorRunResult.kt#L107)). Including an `exception` would cause unit tests outside the scope of this PR to fail because Exceptions' `.equals()` checks reference equality, not value equality. Fixing this would require a more general effort across the unit tests of all existing Monitor types.

src/main/kotlin/org/opensearch/commons/alerting/model/QueryLevelTriggerRunResult.kt
toepkerd: The PPL side feature that would eliminate the need for this extra field already exists: https://github.com/opensearch-project/sql/issues/3629

src/main/kotlin/org/opensearch/commons/alerting/model/PPLTrigger.kt
lezzago: Ack, have we confirmed with notification plugin that this is how he notification channel id is formatted?

toepkerd: After offline discussion, will be removing this ID check. Channel creation happens in Notifications plugin, and there are no rules/constraints there on the contents of Notification channel IDs. As such, Alerting can't block Monitor creation based on its own constraints for what a Notifications channel ID should follow.

lezzago (approved):

eirsep (approved):

PR #954, opened 2026-05-07
adding copied and reduced versions of SQL plugin classes needed for P…

toepkerd:
### Description
This adds copied, reduced versions of SQL Plugin Request/Response classes to save Alerting plugin the need to depend on SQL plugin.

Original classes being copied:
- https://github.com/opensearch-project/sql/blob/main/plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryRequest.java
- https://github.com/opensearch-project/sql/blob/main/plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryResponse.java
- https://github.com/opensearch-project/sql/blob/main/plugin/src/main/java/org/opensearch/sql/plugin/transport/PPLQueryAction.java

### Check List

### DIFF
--- a/src/main/java/org/opensearch/commons/ppl/PPLQueryAction.java
+++ b/src/main/java/org/opensearch/commons/ppl/PPLQueryAction.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.commons.ppl;
+
+import org.opensearch.action.ActionType;
+
+/**
+ * This is a copied, reduced version of SQL Plugin's PPLQueryAction
+ */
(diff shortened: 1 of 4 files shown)


engechas (approved):

lezzago:
The problem with this is that common-utils is not the source of truth and needs to constantly copy and update from SQL plugin and this can be missed and get stale. 

Also has this been tested locally? I feel there will be serialization and deserialization issues since it is technically the same class.

lezzago:
Please create a github issue here, alerting, SQL plugin to move these transport classes to a common subfolder in SQL plugin repo and publish that to maven. Then common-utils can remove this code and alerting can update its code as well.

lezzago (approved):

PR #956, opened 2026-05-12
Feature/target arn field

eirsep:
Add arn field to Target data class to carry the ARN of the remote resources 
The ARN is parsed by the alerting plugin to extract account ID and resource ID for thread context header propagation.

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/alerting/model/Target.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/model/Target.kt
@@ -14,29 +14,34 @@ import java.io.IOException
  *
  * @property type the target type — LOCAL by default. Extensible for additional target types.
  * @property endpoint the URL of the remote target. Required when type is not LOCAL.
+ * @property arn the ARN of the remote resource. Required when type is not LOCAL.
  */
 data class Target(
     val type: String = LOCAL,
-    val endpoint: String = ""
+    val endpoint: String = "",
+    val arn: String = ""
 ) : Writeable, ToXContent {
 
     init {
         require(type.isNotBlank()) { "target type cannot be empty" }
         if (type != LOCAL) {
             require(endpoint.isNotBlank()) { "endpoint is required when target type is not LOCAL" }
+            require(arn.isNotBlank()) { "arn is required when target type is not LOCAL" }
         }
     }
 
     @Throws(IOException::class)
     constructor(sin: StreamInput) : this(
         type = sin.readString(),
-        endpoint = sin.readString()
+        endpoint = sin.readString(),
+        arn = sin.readString()
     )
 
     override fun toXContent(builder: XContentBuilder, params: ToXContent.Params): XContentBuilder {
         builder.startObject()
             .field(TYPE_FIELD, type)
             .field(ENDPOINT_FIELD, endpoint)
+            .field(ARN_FIELD, arn)
         return builder.endObject()
     }
 
@@ -44,11 +49,13 @@ data class Target(
     override fun writeTo(out: StreamOutput) {
         out.writeString(type)
         out.writeString(endpoint)
+        out.writeString(arn)
     }
 
     companion object {
         const val TYPE_FIELD = "type"
         const val ENDPOINT_FIELD = "endpoint"
+        const val ARN_FIELD = "arn"
         const val LOCAL = "local"
         val DEFAULT = Target()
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/opensearch/commons/alerting/model/Target.kt
```diff
@@ -14,41 +14,48 @@ import java.io.IOException
...
-    val endpoint: String = ""
+    val endpoint: String = "",
+    val arn: String = ""
 ) : Writeable, ToXContent {
 
     init {
         require(type.isNotBlank()) { "target type cannot be empty" }
         if (type != LOCAL) {
             require(endpoint.isNotBlank()) { "endpoint is required when target type is not LOCAL" }
+            require(arn.isNotBlank()) { "arn is required when target type is not LOCAL" }
```
AWSHurneyt, line 29: Should we do any other validation on these fields?

E.g., confirming the endpoint ends with `.amazonaws.com`, and confirming the ARN matches the pattern for collection and domain ARNs?

engechas (dismissed):

src/main/kotlin/org/opensearch/commons/alerting/model/Target.kt
```diff
@@ -14,41 +14,48 @@ import java.io.IOException
...
         }
     }
 
     @Throws(IOException::class)
     constructor(sin: StreamInput) : this(
         type = sin.readString(),
-        endpoint = sin.readString()
+        endpoint = sin.readString(),
```
AWSHurneyt, line 36: Should this be `sin.readOptionalString` if endpoint and arn are only required when `type != LOCAL`?

src/main/kotlin/org/opensearch/commons/alerting/model/Target.kt
```diff
@@ -14,41 +14,48 @@ import java.io.IOException
...
+        arn = sin.readString()
     )
 
     override fun toXContent(builder: XContentBuilder, params: ToXContent.Params): XContentBuilder {
         builder.startObject()
             .field(TYPE_FIELD, type)
             .field(ENDPOINT_FIELD, endpoint)
+            .field(ARN_FIELD, arn)
```
AWSHurneyt, line 44: Similar to the comment in the StreamInput constructor, should we exclude the endpoint and arn fields in the payload when type is LOCAL?

src/main/kotlin/org/opensearch/commons/alerting/model/Target.kt
eirsep: will add the arn validation in the Create monitor flow in Alerting plugin and that way i can avoid introducing aws sdk v2 dependency in Common utils

src/main/kotlin/org/opensearch/commons/alerting/model/Target.kt
eirsep: when type is LOCAL we are not mandating that the field is null so i dont think we should exclude it ourselves.

engechas (dismissed):

src/main/kotlin/org/opensearch/commons/alerting/model/Target.kt
eirsep: leaving it as readString for both arn and endpoint as we have defined them as `String` type variables and not `String?`

eirsep:
```
./gradlew build > build.log 2>&1 && echo -e "\n\n=== BUILD SUMMARY (LAST 20 LINES) ===\n" && tail -n 20 build.log (using tool: shell)



=== BUILD SUMMARY (LAST 20 LINES) ===


SecureClientWrapperTests > test no tenant id header does not inject null() PASSED

TenantContextTests > test currentTenantId returns null for single tenant deployment() PASSED

TenantContextTests > test currentTenantId returns tenant id when set() PASSED

TenantContextTests > test currentTenantId returns null when tenant id header is absent() PASSED

> Task :spotlessInternalRegisterDependencies UP-TO-DATE
> Task :spotlessJava UP-TO-DATE
> Task :spotlessJavaCheck UP-TO-DATE
> Task :spotlessCheck UP-TO-DATE
> Task :jacocoTestReport
> Task :check
> Task :build

AWSHurneyt (approved):

vikhy-aws (approved):

engechas (approved):

PR #968, opened 2026-06-09
Fix ClassCastException from immutable map returned by StreamInput.readMap()

### LINKED ISSUE #967: Fix ClassCastException from immutable map returned by StreamInput.readMap()
## Summary

`StreamInput.readMap()` documents that for zero-size maps it *might* return an immutable map (`Collections.emptyMap()`). Three deserialization constructors in this repo perform an unchecked cast (`suppressWarning`) that assumes a `MutableMap`, causing a `ClassCastException` at runtime when an empty map is deserialized over transport.

**Exception:**
```
kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap
```

**Affected files:**
- `Monitor.kt` — `uiMetadata = suppressWarning(sin.readMap())`
- `MonitorMetadata.kt` — `lastRunContext = Monitor.suppressWarning(sin.readMap())`
- `Alert.kt` — `queryResults = sin.readList { input -> suppressWarning(input.readMap()) }`

## Root cause

`StreamOutput.writeMap` / `StreamInput.readMap` guarantee only that the result implements `java.util.Map`. The Javadoc states:

> If the returned map contains any entries it will be mutable. If it is empty it might be immutable.

The `suppressWarning()` helper performs `map as MutableMap<String, Any>` without a defensive copy, which fails when the deserialized map is `Collections.emptyMap()`.

This was identified during review of opensearch-project/security-analytics#1722, where `TransportIndexDetectorAction` passes `Map.of()` (empty immutable Java map) as `uiMetadata` when constructing a `Monitor`. That gets serialized over transport as a zero-size map, `readMap()` returns `Collections.emptyMap()`, and the cast fails.

See the root-cause analysis comment: https://github.com/opensearch-project/security-analytics/pull/1722#issuecomment-2594707019

Related SAP internal tracking: https://github.wdf.sap.corp/PlusOne/greenhouse/pull/1898

## Fix

Replace `suppressWarning(sin.readMap())` with `sin.readMap()?.toMutableMap() ?: mutableMapOf()` at all three callsites. This eliminates the unchecked cast and guarantees a mutable map regardless of what `readMap()` returns.

## Related

- opensearch-project/security-analytics#1722
- opensearch-project/security-analytics#1715
- opensearch-project/security-analytics#1553

thecodingshrimp:
## Summary

Fixes #967

`StreamInput.readMap()` documents that for zero-size maps it *might* return an immutable map (`Collections.emptyMap()`). Three deserialization constructors in this repo perform an unchecked cast (`suppressWarning`) that assumes a `MutableMap`, causing a `ClassCastException` at runtime when a monitor with empty `uiMetadata`, `lastRunContext`, or `queryResults` is deserialized over transport.

**Exception:**
```
kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap
```

## Root cause

`StreamOutput.writeMap` / `StreamInput.readMap` guarantee only that the result implements `java.util.Map`. The Javadoc states:

> If the returned map contains any entries it will be mutable. If it is empty it might be immutable.

The `suppressWarning()` helper performs `map as MutableMap<String, Any>` without a defensive copy, which fails when the deserialized map is `Collections.emptyMap()`.

Identified via opensearch-project/security-analytics#1722 — `TransportIndexDetectorAction` passes `Map.of()` as `uiMetadata` when constructing a `Monitor`, which serializes as a zero-size map, `readMap()` returns `Collections.emptyMap()`, and the cast fails.

## Changes

Replace `suppressWarning(sin.readMap())` with `sin.readMap()?.toMutableMap() ?: mutableMapOf()` at all three callsites and remove the `suppressWarning()` helper function entirely.

| File | Change |
|------|--------|
| `Monitor.kt` | `uiMetadata` deserialization — safe copy |
| `MonitorMetadata.kt` | `lastRunContext` deserialization — safe copy |
| `Alert.kt` | `queryResults` deserialization — safe copy; remove `suppressWarning` import |

## Testing

All existing Monitor-related tests pass (`./gradlew test --tests "*Monitor*"` → BUILD SUCCESSFUL).

## Check List

- [x] Existing tests pass
- [x] Commits signed off (DCO)
- [ ] New tests (not added — fix is a defensive copy at deserialization; behaviour is identical for non-empty maps)

## Related

- Fixes #967
- opensearch-project/security-analytics#1722
- opensearch-project/security-analytics#1715
- opensearch-project/security-analytics#1553

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/alerting/action/DocLevelMonitorFanOutResponse.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/action/DocLevelMonitorFanOutResponse.kt
@@ -8,6 +8,7 @@ package org.opensearch.commons.alerting.action
 import org.opensearch.commons.alerting.model.DocumentLevelTriggerRunResult
 import org.opensearch.commons.alerting.model.InputRunResults
 import org.opensearch.commons.alerting.util.AlertingException
+import org.opensearch.commons.alerting.util.readMapAsMutableMap
 import org.opensearch.core.action.ActionResponse
 import org.opensearch.core.common.io.stream.StreamInput
 import org.opensearch.core.common.io.stream.StreamOutput
@@ -30,9 +31,9 @@ class DocLevelMonitorFanOutResponse : ActionResponse, ToXContentObject {
         nodeId = sin.readString(),
         executionId = sin.readString(),
         monitorId = sin.readString(),
-        lastRunContexts = sin.readMap()!! as MutableMap<String, Any>,
+        lastRunContexts = sin.readMapAsMutableMap() as MutableMap<String, Any>,
         inputResults = InputRunResults.readFrom(sin),
-        triggerResults = suppressWarning(sin.readMap(StreamInput::readString, DocumentLevelTriggerRunResult::readFrom)),
+        triggerResults = readTriggerResults(sin),
         exception = sin.readException()
     )
 
@@ -84,9 +85,11 @@ class DocLevelMonitorFanOutResponse : ActionResponse, ToXContentObject {
     }
 
     companion object {
-        @Suppress("UNCHECKED_CAST")
-        fun suppressWarning(map: MutableMap<String?, Any?>?): Map<String, DocumentLevelTriggerRunResult> {
-            return map as Map<String, DocumentLevelTriggerRunResult>
+        private fun readTriggerResults(sin: StreamInput): Map<String, DocumentLevelTriggerRunResult> {
+            val raw = sin.readMap(StreamInput::readString, DocumentLevelTriggerRunResult::readFrom)
+            if (raw.isEmpty()) return mutableMapOf()
+            @Suppress("UNCHECKED_CAST")
+            return HashMap(raw as Map<String, DocumentLevelTriggerRunResult>)
         }
     }
 }
--- a/src/main/kotlin/org/opensearch/commons/alerting/aggregation/bucketselectorext/BucketSelectorExtAggregationBuilder.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/aggregation/bucketselectorext/BucketSelectorExtAggregationBuilder.kt
@@ -42,7 +42,8 @@ class BucketSelectorExtAggregationBuilder :
     @Throws(IOException::class)
     @Suppress("UNCHECKED_CAST")
     constructor(sin: StreamInput) : super(sin, NAME.preferredName) {
-        bucketsPathsMap = sin.readMap() as MutableMap<String, String>
+        @Suppress("UNCHECKED_CAST")
+        bucketsPathsMap = (sin.readMap()?.toMutableMap() ?: mutableMapOf()) as Map<String, String>
         script = Script(sin)
         gapPolicy = BucketHelpers.GapPolicy.readFrom(sin)
         parentBucketPath = sin.readString()
--- a/src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt
@@ -3,11 +3,11 @@ package org.opensearch.commons.alerting.model
 import org.opensearch.Version
 import org.opensearch.common.lucene.uid.Versions
 import org.opensearch.commons.alerting.alerts.AlertError
-import org.opensearch.commons.alerting.model.Monitor.Companion.suppressWarning
 import org.opensearch.commons.alerting.util.IndexUtils.Companion.NO_SCHEMA_VERSION
 import org.opensearch.commons.alerting.util.instant
 import org.opensearch.commons.alerting.util.optionalTimeField
 import org.opensearch.commons.alerting.util.optionalUserField
+import org.opensearch.commons.alerting.util.readMapAsMutableMap
 import org.opensearch.commons.authuser.User
 import org.opensearch.core.common.io.stream.StreamInput
 import org.opensearch.core.common.io.stream.StreamOutput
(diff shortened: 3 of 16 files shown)


thecodingshrimp:
Thanks for the automated review. Addressing each flagged item (via claude):

### Unsafe cast: \`WorkflowRunResult.kt\` → \`Map<String, ChainedAlertTriggerRunResult>\`

The bot suggested switching to the typed \`readMap(StreamInput::readString, ChainedAlertTriggerRunResult::readFrom)\` overload to avoid the unchecked cast.

This is not viable without also changing the write side — the existing \`out.writeMap(triggerResults)\` uses the generic (type-tagged) wire format, while the typed \`readMap\` overload expects the type-free format written by \`out.writeMap(map, keyWriter, valueWriter)\`. Changing both sides in the same PR would introduce a **wire-format break** for rolling upgrades. The wire format for this map was established in its original introduction; preserving it is necessary.

What I've done instead (commit 9e0e029): added the missing \`@Suppress("UNCHECKED_CAST")\` on the constructor, which was inadvertently omitted. The cast is safe because the values were serialized as \`ChainedAlertTriggerRunResult\` instances by \`writeTo()\`.

### Unsafe cast: \`MonitorMetadata.kt\` → \`MutableMap<String, String>\`

The suggestion to add per-entry type validation via \`as? String ?: throw IllegalStateException\` is over-defensive. The data at this position was always written by \`MonitorMetadata.writeTo()\` which casts \`sourceToQueryIndexMapping: MutableMap<String, String>\` down to \`MutableMap<String, Any>\` for the generic writer. The types are guaranteed by the write side. Adding entry-level type assertions would obscure a straightforward deserialization with complexity that adds no real value.

### Unsafe cast: \`DocLevelMonitorFanOutResponse.kt\` → \`Map<String, DocumentLevelTriggerRunResult>\`

The suggestion to replace the cast with \`.mapValues { it.value as? DocumentLevelTriggerRunResult ?: throw ... }\` is redundant: \`readMap(StreamInput::readString, DocumentLevelTriggerRunResult::readFrom)\` already guarantees each value was deserialized via \`DocumentLevelTriggerRunResult::readFrom\`. A subsequent runtime check on what the typed deserializer just produced adds only overhead.

In summary: the flagged casts are all inherently unchecked due to JVM type erasure, but each is provably safe given the matching write side. The only actionable item was the missing \`@Suppress\` annotation in \`WorkflowRunResult\`.

thecodingshrimp:
Added regression tests in  (commit 504f10c) covering all six fixed call sites. Each test exercises the empty-map path: serializes via `writeTo()`, deserializes via the `StreamInput` constructor, and asserts the round-trip succeeds and the result map is mutable.

src/main/kotlin/org/opensearch/commons/alerting/action/DocLevelMonitorFanOutResponse.kt
```diff
@@ -25,14 +25,16 @@ class DocLevelMonitorFanOutResponse : ActionResponse, ToXContentObject {
...
     val exception: AlertingException?
 
+    @Suppress("UNCHECKED_CAST")
     @Throws(IOException::class)
     constructor(sin: StreamInput) : this(
         nodeId = sin.readString(),
         executionId = sin.readString(),
         monitorId = sin.readString(),
-        lastRunContexts = sin.readMap()!! as MutableMap<String, Any>,
+        lastRunContexts = sin.readMap()?.toMutableMap() ?: mutableMapOf(),
```
andrross: @thecodingshrimp  What do you think about using kotlin extensions here? Something like:

```java
package org.opensearch.commons.alerting.util

import org.opensearch.core.common.io.stream.StreamInput
import org.opensearch.core.common.io.stream.Writeable

/**
 * Reads a map written by `StreamOutput.writeMap` and always returns a mutable map.
 *
 * Why this exists:
 * `StreamInput.readMap()` only guarantees the result implements `java.util.Map`. Its
 * Javadoc states that a non-empty result will be mutable, but an empty result *might*
 * be immutable (`Collections.emptyMap()`). This extension guarantees the returned
 * map is mutable.
 */
@Suppress("UNCHECKED_CAST")
fun StreamInput.readMapAsMutableMap(): MutableMap<String, Any> {
    val map = this.readMap() ?: return mutableMapOf()
    return if (map is MutableMap<*, *>) {
        map as MutableMap<String, Any>
    } else {
        // Immutable (e.g. Collections.emptyMap()) or unknown type: copy every entry
        // into a fresh mutable map.
        LinkedHashMap(map) as MutableMap<String, Any>
    }
}

/**
 * Typed variant for maps written with explicit key/value readers
 * (`StreamOutput.writeMap(map, keyWriter, valueWriter)`).
 */
fun <K, V> StreamInput.readMapAsMutableMap(
    keyReader: Writeable.Reader<K>,
    valueReader: Writeable.Reader<V>
): MutableMap<K, V> {
    val map = this.readMap(keyReader, valueReader) ?: return mutableMapOf()
    return if (map is MutableMap<K, V>) {
        map
    } else {
        LinkedHashMap(map)
    }
}
``` 

Then all the call sites that need a mutable map from StreamInput can use one of these methods.

thecodingshrimp: Thanks for the suggestion — agreed that centralizing this in an extension is the right call. I went with a cast-free version to avoid introducing a new `@Suppress`:

```kotlin
fun StreamInput.readMapAsMutableMap(): MutableMap<String, Any> =
    readMap()?.toMutableMap() ?: mutableMapOf()
```

This is placed in `util/StreamInputExtensions.kt` and used across all the fixed call sites. The `is MutableMap<*, *>` branch check in your proposal avoids an unnecessary copy for non-empty maps, but `toMutableMap()` is safe and keeps the helper annotation-free. I also extended the fix to cover the remaining `suppressWarning(sin.readMap())` calls across the `TriggerRunResult` family and deleted the now-dead `suppressWarning` helpers.

hunter007861 (approved):

hunter007861 (approved):

PR #981, opened 2026-07-22
Override DocRequest.type() on alerting request classes; tolerate ancillary top-level fields in ScheduledJob.parse

DarshitChanpura:
## Summary

Two related changes required by alerting's onboarding to the security plugin's resource-sharing framework (RSC), driven by [opensearch-project/alerting#2180](https://github.com/opensearch-project/alerting/pull/2180) and [opensearch-project/security#6323](https://github.com/opensearch-project/security/pull/6323).

### 1. `DocRequest.type()` overrides

The security plugin's [ResourceAccessEvaluator](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java#L134) gates a request only when `DocRequest.type()` is contained in the configured `protected_types` list. The default `type()` returns `\"indices\"`, so without an override the evaluator skipped every alerting request and the ActionFilter granted access unconditionally.

Follow-up to #980 which added `DocRequest` but did not override `type()`.

- Add `MONITOR_RESOURCE_TYPE` and `WORKFLOW_RESOURCE_TYPE` constants to `AlertingConstants`.
- Monitor-scoped request classes (`IndexMonitorRequest`, `DeleteMonitorRequest`, `GetMonitorRequest`, `AcknowledgeAlertRequest`, `AcknowledgeChainedAlertRequest`, `GetAlertsRequest`, `GetFindingsRequest`, `DocLevelMonitorFanOutRequest`) return `\"monitor\"`.
- Workflow-scoped request classes (`IndexWorkflowRequest`, `DeleteWorkflowRequest`, `GetWorkflowRequest`, `GetWorkflowAlertsRequest`) return `\"workflow\"`.
- Comment requests unchanged — comments are not a protected resource type.

### 2. `ScheduledJob.parse` tolerates ancillary top-level fields

When a doc is stored on a resource-sharing-protected index, the security plugin's `postIndex` hook writes `all_shared_principals` as a top-level field on the doc for DLS filtering. Legacy stored docs may also carry other top-level fields written by admin migration tools. The old `parse` implementation assumed the doc had a single wrapper (`{\"monitor\": {...}}` or `{\"workflow\": {...}}`) and threw on any additional top-level fields.

Both `parse` overloads now iterate all top-level fields, dispatch to the registered subtype parser (`Monitor.parse` / `Workflow.parse`) when the field name matches a known wrapper, and skip any other top-level fields via `skipChildren()`.

## Backward compatibility

- Old-shape docs (`{\"monitor\": {...}}`) parse identically — the wrapper is still the only recognized field.
- New-shape docs (`{\"monitor\": {...}, \"all_shared_principals\": [...]}`) now parse instead of throwing.
- No changes to Monitor/Workflow XContent output — the write shape is unchanged, so downstream consumers see the same docs they always have.

## Related

- opensearch-project/alerting#2180
- opensearch-project/security#6018 (defined the `protected_types` gating)
- opensearch-project/security#6323 (multi-provider `typeField` iteration + per-provider owner paths; consumes the workflow type value)

### DIFF
--- a/src/main/kotlin/org/opensearch/commons/alerting/action/DeleteWorkflowRequest.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/action/DeleteWorkflowRequest.kt
@@ -4,6 +4,7 @@ import org.opensearch.action.ActionRequest
 import org.opensearch.action.ActionRequestValidationException
 import org.opensearch.action.DocRequest
 import org.opensearch.commons.alerting.model.ScheduledJob
+import org.opensearch.commons.alerting.util.AlertingConstants
 import org.opensearch.core.common.io.stream.StreamInput
 import org.opensearch.core.common.io.stream.StreamOutput
 import java.io.IOException
@@ -45,5 +46,9 @@ class DeleteWorkflowRequest : ActionRequest, DocRequest {
 
     override fun id(): String? {
         return workflowId
+    }
+
+    override fun type(): String {
+        return AlertingConstants.MONITOR_RESOURCE_TYPE
     }
 }
--- a/src/main/kotlin/org/opensearch/commons/alerting/action/DeleteWorkflowRequest.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/action/DeleteWorkflowRequest.kt
@@ -49,6 +49,6 @@ class DeleteWorkflowRequest : ActionRequest, DocRequest {
     }
 
     override fun type(): String {
-        return AlertingConstants.MONITOR_RESOURCE_TYPE
+        return AlertingConstants.WORKFLOW_RESOURCE_TYPE
     }
 }
--- a/src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
@@ -206,13 +206,6 @@ data class Monitor(
 
     private fun createXContentBuilder(builder: XContentBuilder, params: ToXContent.Params, secure: Boolean): XContentBuilder {
         builder.startObject()
-        // `with_resource_type` is orthogonal to `with_type` and is meant for the on-disk representation
-        // in `SCHEDULED_JOBS_INDEX` so the security plugin's resource-sharing framework can discriminate
-        // monitor vs workflow docs via a single indexed field. Storage writes opt in explicitly; IPC /
-        // REST payloads keep the historical shape to preserve cross-version compatibility.
-        if (params.paramAsBoolean("with_resource_type", false)) {
-            builder.field(ScheduledJob.RESOURCE_TYPE_FIELD, type)
-        }
         if (params.paramAsBoolean("with_type", false)) builder.startObject(type)
         builder.field(TYPE_FIELD, type)
             .field(SCHEMA_VERSION_FIELD, schemaVersion)
--- a/src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
+++ b/src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
@@ -206,10 +206,14 @@ data class Monitor(
 
     private fun createXContentBuilder(builder: XContentBuilder, params: ToXContent.Params, secure: Boolean): XContentBuilder {
         builder.startObject()
-        if (params.paramAsBoolean("with_type", false)) {
+        // `with_resource_type` is orthogonal to `with_type` and is meant for the on-disk representation
+        // in `SCHEDULED_JOBS_INDEX` so the security plugin's resource-sharing framework can discriminate
+        // monitor vs workflow docs via a single indexed field. Storage writes opt in explicitly; IPC /
+        // REST payloads keep the historical shape to preserve cross-version compatibility.
+        if (params.paramAsBoolean("with_resource_type", false)) {
             builder.field(ScheduledJob.RESOURCE_TYPE_FIELD, type)
-            builder.startObject(type)
         }
+        if (params.paramAsBoolean("with_type", false)) builder.startObject(type)
         builder.field(TYPE_FIELD, type)
             .field(SCHEMA_VERSION_FIELD, schemaVersion)
             .field(NAME_FIELD, name)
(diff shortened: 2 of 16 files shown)


src/main/kotlin/org/opensearch/commons/alerting/action/DeleteWorkflowRequest.kt
```diff
@@ -46,4 +47,8 @@ class DeleteWorkflowRequest : ActionRequest, DocRequest {
     override fun id(): String? {
         return workflowId
     }
+
+    override fun type(): String {
+        return AlertingConstants.MONITOR_RESOURCE_TYPE
```
cwperks: Is workflow not a distinct type?

cwperks (approved):

src/main/kotlin/org/opensearch/commons/alerting/model/Workflow.kt
```diff
@@ -106,7 +106,10 @@ data class Workflow(
         secure: Boolean
     ): XContentBuilder {
         builder.startObject()
-        if (params.paramAsBoolean("with_type", false)) builder.startObject(type)
+        if (params.paramAsBoolean("with_type", false)) {
+            builder.field(ScheduledJob.RESOURCE_TYPE_FIELD, type)
```
cwperks: Could this be breaking?

src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt
```diff
@@ -206,7 +206,10 @@ data class Monitor(
 
     private fun createXContentBuilder(builder: XContentBuilder, params: ToXContent.Params, secure: Boolean): XContentBuilder {
         builder.startObject()
-        if (params.paramAsBoolean("with_type", false)) builder.startObject(type)
+        if (params.paramAsBoolean("with_type", false)) {
+            builder.field(ScheduledJob.RESOURCE_TYPE_FIELD, type)
```
cwperks: Same here.

src/main/kotlin/org/opensearch/commons/alerting/model/ScheduledJob.kt
```diff
@@ -51,7 +75,18 @@ interface ScheduledJob : BaseModel {
...
+            // scratch fields).
+            var next = xcp.nextToken()
+            while (next != null && next != XContentParser.Token.END_OBJECT) {
+                if (next == XContentParser.Token.FIELD_NAME) {
+                    xcp.nextToken()
+                    xcp.skipChildren()
+                }
+                next = xcp.nextToken()
+            }
+            XContentParserUtils.ensureExpectedToken(XContentParser.Token.END_OBJECT, next, xcp)
```
DarshitChanpura: Why are these changes needed?

**What the file is for**

`.opendistro-alerting-config` is a shared filing cabinet. Monitors and workflows both live there. Each doc looks like a folder with one big envelope inside:

- Monitor doc: `{ "monitor":  { …monitor fields… } }`
- Workflow doc: `{ "workflow": { …workflow fields… } }`

`ScheduledJob.parse` is the mail-opener. Given one of these docs, it opens the outer folder, looks at the envelope's name (`monitor` or `workflow`), and hands the envelope's contents to the matching parser (`Monitor.parse` or `Workflow.parse`).

**What the old parser did**

The old code was hard-coded to a rigid procedure:

1. Open the outer folder (`{`)
2. Expect exactly ONE thing inside — the envelope
3. Read the envelope's name
4. Open the envelope, hand contents to the specialist
5. Expect the outer folder to close (`}`) with nothing else inside

If it found anything other than the single envelope — for example, a sticky note stuck to the inside of the folder — it errored out.

**Why that broke**

When resource-sharing is on, the security plugin stamps every doc with an `all_shared_principals: [...]` field for DLS. That's a sticky note next to the envelope. The old parser saw two things in the folder and threw.

**What the new parser does**

Both overloads (`parse(xcp, id, version)` and `parse(xcp, type, id, version)`) now delegate to a single private helper `parseWrapper`. It reads every item inside the outer folder:

- If the item's name is `monitor` or `workflow` (defined in `SCHEDULED_JOB_WRAPPER_FIELDS`), that's the envelope — open it and hand off to the specialist. The 2-arg overload additionally requires the name to match the caller-supplied type.
- Anything else (e.g. `all_shared_principals`) — it's a sticky note, skip via `xcp.skipChildren()`.

At the end it verifies an envelope was found (`requireNotNull(job)`) and that the outer folder closed properly.

The shared helper is **order-independent** — a sticky note can appear before or after the envelope; the wrapper still gets picked up.

src/main/kotlin/org/opensearch/commons/alerting/model/ScheduledJob.kt
```diff
@@ -23,21 +23,45 @@ interface ScheduledJob : BaseModel {
...
-         * This function parses the job, delegating to the specific subtype parser registered in the [XContentParser.getXContentRegistry]
-         * at runtime.  Each concrete job subclass is expected to register a parser in this registry.
-         * The Job's json representation is expected to be of the form:
+         * This function parses the job, delegating to the specific subtype parser registered in
+         * the [XContentParser.getXContentRegistry] at runtime. Each concrete job subclass is
+         * expected to register a parser in this registry. The Job's JSON representation is
+         * expected to be of the form:
          *     { "<job_type>" : { <job fields> } }
+         * Any additional top-level fields (e.g. `all_shared_principals` injected by the security
+         * plugin's resource-sharing framework, or migration scratch fields) are tolerated and
```
cwperks: Can we update this comment if those "scratch fields" were coming from some previous WIP code?

src/main/kotlin/org/opensearch/commons/alerting/model/ScheduledJob.kt
```diff
@@ -51,7 +75,18 @@ interface ScheduledJob : BaseModel {
         fun parse(xcp: XContentParser, type: String, id: String = NO_ID, version: Long = NO_VERSION): ScheduledJob {
```
cwperks: Is there an opportunity for code re-use here if the purpose is similar to the block above? I'm not sure if this is robust to ordering of keys.

cwperks (approved):

PR #993, opened 2026-08-17
Use Eclipse P2 mirror (common-utils)

peterzhuamazon:
Routes Spotless Eclipse P2 downloads through https://ci.opensearch.org/ to avoid direct download.eclipse.org outages.

Tracking bulletin: https://github.com/opensearch-project/opensearch-build/issues/6421#issuecomment-5271186726


peterzhuamazon:
expected mirror update.

.github/workflows/auto-release.yml
```diff
@@ -1,29 +0,0 @@
-name: Releases
```
gaiksaya, line 1: We need the auto-release workflow, maybe similar to how other standalone repos have. We can remove the token usage as it is not required.

